PR #50, opened 2014-07-11
Enable completion for packages, patterns, prodcuts, updates, and patches

Aikhjarto:
Implemented the TODO entry in zypper's bash completion and added some other completion feature.


bmwiedemann:
I did not know about this PR, so I did #66 by myself, which includes some caching to avoid the multi-second delay from zypper calls.

mlandres:
Thanks Aikhjarto, I'm closing this in favor of #66. I hope we included all of your feature there.

PR #60, opened 2015-03-26
Aptitude updates

bmwiedemann:
update our aptitude compat layer
to also be a drop-in for apt-get
and to allow the mapping of package-names

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,3 +94,7 @@ INSTALL(
   PROGRAMS tools/aptitude
   DESTINATION ${INSTALL_PREFIX}/bin
 )
+INSTALL(
+  FILES tools/apt-packagemap.d/10-packagemap.pm tools/apt-packagemap.d/50-libperl.pm tools/apt-packagemap.d/50-libruby.pm tools/apt-packagemap.d/90-devel.pm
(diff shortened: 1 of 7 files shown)


KAMiKAZOW:
What would be the best approach for package name mapping under non-openSUSE distros?
One workaround would be to patch the apt-packagemap.d/* files while packaging. It's doable but also very inelegant.
I'd prefer some way to maintain the list here upstream. In 10-packagemap.pm roughly half of the remaps need to be changed for Fedora: https://github.com/openSUSE/zypper/compare/master...KAMiKAZOW:patch-1

bmwiedemann:
Ideally, we should have a central place where we store package-name-mappings (e.g. AppStream?) and generate these files from the database.

For now, it would be nicer for you to have a 2nd block with just the differences, like this
```my %fedorapkgmap = qw(python-mysqldb MySQL-python ...);

if(os_is "fedora") {
   foreach my $k (keys(%fedorapkgmap)) { $pkgmap{$k}=$fedorapkgmap{$k} }
}```

With the os_is check still to be done.

bmwiedemann:
@KAMiKAZOW  #65 should make this work. Can you test it?

KAMiKAZOW:
Give me until tomorrow or the day after. I'm currently fighting allergies and it's supposed to rain to coming days.

PR #62, opened 2015-04-07
move duplicated code into wcwidth_without_ctrlseq function

bmwiedemann:


src/utils/text.cc
```diff
@@ -171,14 +170,7 @@ void mbs_write_wrapped(ostream & out, const string & text,
     bytes_read = mbrtowc (&wc, s, s_bytes, &shift_state);
     if (bytes_read > 0)
```
bmwiedemann: note (unrelated to patch), that the special -2 and -1 cases are not handled here - and this is an unsigned int, so it will still go through the if part.

bmwiedemann:
updated to use namespace (though I find it less straightforward than the classic static inline used everywhere in C code)
also added a 2nd commit that still needs careful testing

src/utils/text.cc
mlandres: Looking at the history, I'd say the author simply forgot that ```bytes_read``` is unsigned and wanted to catch the case in the trailing else. I'd simply change to:
```
if ( bytes_read > 0 && bytes_read < (size_t) -2 )

```

PR #65, opened 2015-04-24
Aptitude fedora support

bmwiedemann:
add generic python package name mapping
have per-OS packagemaps to allow this to work on Fedora


bmwiedemann:
did a quick test using
```bash
yum install yum-utils
yum-config-manager --add-repo http://download.opensuse.org/repositories/home:/KAMiKAZOW:/Fedora/Fedora_21/
yum install --nogpgcheck zypper zypper-aptitude
```

PR #66, opened 2015-04-26
add bash completion for install+remove+update commands (bnc#928650)

bmwiedemann:
https://bugzilla.opensuse.org/show_bug.cgi?id=928650

### DIFF
--- a/src/bash-completion.sh
+++ b/src/bash-completion.sh
@@ -112,9 +112,12 @@ _zypper() {
 	fi
 
 	if test -n "$command" ; then
-      		opts=$(LC_ALL=POSIX $ZYPPER -q help $command 2>&1 | sed -e "1,/$magic_string/d" -e 's/.*--/--/' -e 's/ .*//')
+		if ! [[ $cur =~ ^[^-] ]] ; then
+			opts=$(LC_ALL=POSIX $ZYPPER -q help $command 2>&1 | sed -e "1,/$magic_string/d" -e 's/.*--/--/' -e 's/ .*//')
+		fi
 
 		#handling individual commands if they need more then we can dig from help
+		if ! [[ $cur =~ ^- ]] ; then
 		case "$command" in
 			help | \?)
 				opts=(${ZYPPER_CMDLIST[@]})
@@ -147,6 +150,7 @@ _zypper() {
 					}'))
 			;;
 		esac
+		fi
 		IFS=$'\n'
 		COMPREPLY=($(compgen -W "${opts[*]}" -- ${cur}))
 		_strip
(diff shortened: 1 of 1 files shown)


bmwiedemann:
After a discussion, we concluded that it would be best, if libzypp or repo2solv could produce a per-repo cache file after a refresh. The cache file would contain such lines:
```
bash-4.2-75.3.1.x86_64
rpm-4.11.3-12.1.x86_64
```

mlandres:
Can we include ```{patch,pattern,product}-info``` commands to provide a result similar to #50?

bmwiedemann:
added those 3 ... they are somewhat slow though (and I dont want to spend time optimizing with caches if we are changing this part soon)

mlandres:
libzypp-15.1.3 (available in obs://zypp:HEAD) supports writing solv.idx files to speed up completion.

src/bash-completion.sh
```diff
@@ -23,8 +22,46 @@ _strip()
...
 
+_get_idx_files()
+{
+	local d=/var/cache/zypp/solv
+	local idx=$d/*/solv.idx
+	# this part covers the case
+	# where no idx files have been created by libzypp yet
+	# TODO: drop in 2019
+        set +o noglob
+	if echo $idx | grep -q -F '*' ; then
```
bmwiedemann: This part seems to be a bit fragile as there might have been zypper ref calls before the first completion. Also only works for root user.

src/bash-completion.sh
```diff
@@ -23,8 +22,46 @@ _strip()
...
 
+_get_idx_files()
+{
+	local d=/var/cache/zypp/solv
+	local idx=$d/*/solv.idx
+	# this part covers the case
+	# where no idx files have been created by libzypp yet
+	# TODO: drop in 2019
+	set +o noglob
+	if echo $idx | grep -q -F '*' ; then
```
bmwiedemann: This part seems to be a bit fragile as there might have been zypper ref calls before the first completion. Also only works for root user.

src/bash-completion.sh
mlandres: You can't silently remove the solv files.
Existing solv files may belong to volatile media (CD/DVD/USB), so you'd enforce zypper asking to insert the media again. I can try to create missing idx files in zypp even if the solv file is up-to-date.

src/bash-completion.sh
bmwiedemann: With https://github.com/openSUSE/libzypp/pull/44 we could drop this ugly workaround.

bmwiedemann:
Updated the code to not have any ugly workarounds anymore. IMHO it is ready to be merged.

PR #69, opened 2015-05-19
avoid blocking+malloc below signal_handler (bnc#929826)

bmwiedemann:
the old code would deadlock and block until killed with kill -9

https://bugzilla.opensuse.org/show_bug.cgi?id=929826

see also
http://pubs.opengroup.org/onlinepubs/009695399/functions/xsh_chap02_04.html#tag_02_04_03
for a list of signal-safe functions

### DIFF
--- a/src/Zypper.cc
+++ b/src/Zypper.cc
@@ -149,8 +149,6 @@ Zypper::Ptr & Zypper::instance()
 
   if (!_instance)
     _instance.reset(new Zypper());
-  else
-    XXX << "Got an existing instance." << endl;
 
   return _instance;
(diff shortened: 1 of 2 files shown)


bmwiedemann:
note: some segfaults still left

mlandres:
The only clean solution is leaving the exit handler and exiting ASAP. Appdata trigger, cleanup of temp. repos, directories, etc.. must be performed.

mlandres:
Should be resolved meanwhile. The exit handler is left ASAP and should not perform any cleanup anymore.

PR #75, opened 2015-07-13
show installed packages in Summary when --no-build-deps is used for source-install

tansteventan:


src/Summary.cc
```diff
@@ -137,6 +137,13 @@ void Summary::readPool(const zypp::ResPool & pool)
     }
   }
 
+  Zypper& zypper = *Zypper::instance();
+  for_(it, zypper.runtimeData().srcpkgs_to_install.begin(), zypper.runtimeData().srcpkgs_to_install.end())
+  {
+    _toinstall[ResKind::srcpackage].insert(ResPair(nullptr, *it));
+    to_be_installed[ResKind::srcpackage].insert(*it);
+  }
+
```
mlandres: Thanks for the fix.

- The  ```_toinstall``` insertion here should be superfluous as the loop below (iterating ```to_be_installed```) does it as well..

- A few lines below your change there is a comment ```/* This will work again after commit refactoring: all the srcpackages will ....*/```. Please remove it, as your fix makes it obsolete.

- (minor) In new code I'd prefer this form 
```
for ( const auto & spkg : Zypper::instance()->runtimeData().srcpkgs_to_install )
{ to_be_installed[ResKind::srcpackage].insert(spkg); }
```

tansteventan:
@mlandres Thanks for the suggestion. I have updated the code and hopefully this can be merged because I use the --no-build-deps very often :)

PR #81, opened 2015-10-28
Provide priority option on addrepo command.

boombatower:
- refactor priority parsing into parse_priority()
- add priority to addrepo for both by repo file and url cases

I have desired this feature for quite a while, since it seems the only proper way to add repositories is with a priority, but that requires adding then modifying. That makes documentation on how to add a repo cumbersome in addition to usage.

This has been discussed on mailing list a few times including http://lists.opensuse.org/opensuse-factory/2015-07/msg00377.html thread.

Semi-related it seem like a good idea for repositories on obs to include a priority in their `.repo` file relevant to their intended usage. Most likely this would be best done with string priorities as suggested in the todo in `repos.cc`. I may implement that next after this pull request has been accepted.

### DIFF
--- a/src/repos.cc
+++ b/src/repos.cc
@@ -70,6 +70,27 @@ void safe_lexical_cast (Source s, Target &tr) {
   }
 }
 
+long long parse_priority(Zypper & zypper) {
+  long long prio = 0;
+  parsed_opts::const_iterator tmp1;
+  if ((tmp1 = zypper.cOpts().find("priority")) != zypper.cOpts().end())
+  {
+    //! \todo use some preset priorities (high, medium, low, ...)
+
+    string prio_str = *tmp1->second.begin();
+    safe_lexical_cast(prio_str, prio); // try to make an int out of the string
+    if (prio < 1)
+    {
+      zypper.out().error(boost::str(format(
+        _("Invalid priority '%s'. Use a positive integer number. The greater the number, the lower the priority."))
+        % prio_str));
+      zypper.setExitCode(ZYPPER_EXIT_ERR_INVALID_ARGS);
+    }
+  }
+
+  return prio;
+}
+
 // | Enabled | GPG Check |  Colored strings for enabled and GPG Check status
 // +---------+-----------+
 // | Yes     | (  ) No   |
@@ -1726,6 +1747,8 @@ void add_repo(Zypper & zypper, RepoInfo & repo)
     // translators: property name; short; used like "Name: value"
     p.add( _("GPG Check"), 	repo.gpgCheck() ).paint( ColorContext::MSG_WARNING, repo.gpgCheck() == false );
     // translators: property name; short; used like "Name: value"
+    p.add( _("Priority"),		repo.priority() );
+    // translators: property name; short; used like "Name: value"
     p.add( _("URI"),		repo.baseUrlsBegin(), repo.baseUrlsEnd() );
     s << p;
   }
@@ -1781,6 +1804,11 @@ void add_repo_by_url( Zypper & zypper,
   if (it != zypper.cOpts().end())
     repo.setName(it->second.front());
   repo.addBaseUrl(url);
+
+  long long prio = parse_priority(zypper);
+  if (prio > 1) {
+    repo.setPriority(prio);
+  }
 
   // enable the repo by default
   if ( indeterminate(enabled) )
@@ -1882,6 +1910,11 @@ void add_repo_from_file( Zypper & zypper,
 
     if ( !indeterminate(gpgCheck) )
       repo.setGpgCheck(gpgCheck);
+
+    long long prio = parse_priority(zypper);
+    if (prio > 1) {
+      repo.setPriority(prio);
+    }
 
     MIL << "to-be-added: enabled: " << repo.enabled() << " autorefresh: " << repo.autorefresh() << endl;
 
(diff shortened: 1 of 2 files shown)


src/repos.cc
```diff
@@ -70,6 +70,27 @@ void safe_lexical_cast (Source s, Target &tr) {
...
+    //! \todo use some preset priorities (high, medium, low, ...)
+
+    string prio_str = *tmp1->second.begin();
+    safe_lexical_cast(prio_str, prio); // try to make an int out of the string
+    if (prio < 1)
+    {
+      zypper.out().error(boost::str(format(
+        _("Invalid priority '%s'. Use a positive integer number. The greater the number, the lower the priority."))
+        % prio_str));
+      zypper.setExitCode(ZYPPER_EXIT_ERR_INVALID_ARGS);
```
mlandres: I know,  the original code used ```long long```, but:
* repo priority should be ```unsigned parse_priority( Zypper & zypper )``` as this is what zypp:RepoInfo uses.
* __most important:__ if the cli arg is invalid (negative or no number) zypper must not continue! Add ```ZYPP_THROW(ExitRequestException("Invalid priority."));``` after ```setExitCode(...);```
* ```-p 0``` is (or should be) a legal priority value, setting the systems default priority: the value returned by ```static unsigned RepoInfo::defaultPriority()```.  
* it must be possible to distinguish between a valid ```-p arg``` and the absence of  ```-p``` on the command line. Thus ```0``` should exclusively be returned if no ```-p``` option is present (>0 for valid priority from CL; else throw).

Maybe something like:

```
unsigned parse_priority( Zypper & zypper )
{
  unsigned ret = 0U;
  parsed_opts::const_iterator cArg = zypper.cOpts().find("priority");
  if ( cArg == zypper.cOpts().end() )
    return ret;		// 0: no --priority arg

  int prio = -1;
  std::string prio_str = *cArg->second.begin();
  safe_lexical_cast(prio_str, prio); // try to make an int out of the string

  if ( prio < 0 )
  {
    zypper.out().error(boost::str(format(
      _("Invalid priority '%s'. Use a positive integer number. The greater the number, the lower the priority."))
      % prio_str));
    zypper.setExitCode(ZYPPER_EXIT_ERR_INVALID_ARGS);
    ZYPP_THROW(ExitRequestException("Invalid priority."));
  }

  ret = ( prio ? unsigned(prio) : RepoInfo::defaultPriority() );
  return ret;
}
```

src/repos.cc
```diff
@@ -1782,6 +1805,11 @@ void add_repo_by_url( Zypper & zypper,
     repo.setName(it->second.front());
   repo.addBaseUrl(url);
 
+  long long prio = parse_priority(zypper);
```
mlandres: ```unsigned prio = parse_priority(zypper);``` 
Parsing (as it may throw) should be moved to the beginning of add_repo_by_url; e.g. before ```RepoInfo repo;```.

src/repos.cc
```diff
@@ -1782,6 +1805,11 @@ void add_repo_by_url( Zypper & zypper,
...
+  if (prio > 1) {
+    repo.setPriority(prio);
+  }
```
mlandres:
```
  if ( prio )
    repo.setPriority(prio);
```

src/repos.cc
```diff
@@ -1883,6 +1911,11 @@ void add_repo_from_file( Zypper & zypper,
     if ( !indeterminate(gpgCheck) )
       repo.setGpgCheck(gpgCheck);
 
+    long long prio = parse_priority(zypper);
```
mlandres: like in add_repo_by_url; ```unsigned ...``` and the call should be moved outside the loop; best before ```list<RepoInfo> repos;```.

src/repos.cc
```diff
@@ -1883,6 +1911,11 @@ void add_repo_from_file( Zypper & zypper,
...
+    if (prio > 1) {
+      repo.setPriority(prio);
+    }
+
```
mlandres:
```
  if ( prio )
    repo.setPriority(prio);
```

src/repos.cc
```diff
@@ -2096,23 +2129,8 @@ void modify_repo(Zypper & zypper, const string & alias)
...
-      if (prio < 1)
-      {
-        zypper.out().error(boost::str(format(
-          _("Invalid priority '%s'. Use a positive integer number. The greater the number, the lower the priority."))
-          % prio_str));
-        zypper.setExitCode(ZYPPER_EXIT_ERR_INVALID_ARGS);
-        return;
-      }
-
+    long long prio = parse_priority(zypper);
```
mlandres: ```unsigned...``` and the parser call should be moved up, outside the ```try```-block. No need to setup anything if the cli parsing fails.

src/repos.cc
```diff
@@ -2096,23 +2129,8 @@ void modify_repo(Zypper & zypper, const string & alias)
...
+    if (prio >= 1) {
       if (prio == (int) repo.priority())
```
mlandres:
```
if ( prio )
{
  if ( prio == repo.priority() )
    zypper.out().....
```

boombatower:
I can't help feeling like this should be a followup task as it is unrelated to the feature since all this code was existing.

boombatower:
One item to note on the no-value return of parse_priority(). I debated this in the beginning thinking of using typical `-1` to denote a lack of value, but after noticing the code only allowed `>= 1` meant that `0` was essential `-1`. If you allow setting zero as I understand it is supposed to be valid that represents a change over the current behavior. Secondly, `RepoInfo::defaultPriority()` does not return zero, but instead it returns `99`. If it did return zero that would represent another change since currently repos are added with `99` priority by default. By not setting it at all it does not seem to be written to the file on disk, but if we always return `99` on no value it will start writing it. Does not seem like we need to make that change.

If we want to support a `0` value then I presume we need to return a signed int for simplicity, otherwise since it is not supported now it seems reasonable to leave as is and file an issue if truly desired.

mlandres:
>  I will work on changes never the less.

Thank you.

> ...much of this should be a followup task since all this code was existing

Well, the exception was necessary - the `unsigned` and `-p 0` are indeed followups. The issue caught my eye, but I forgot that it's probably not that obvious to everybody. Sorry. 
Your solution is fine.


(=====
Valid repo priority numbers are `greater than 0`. 
`0` itself is no valid priority value,  but in ZYPP it may always be used to request the `defaultPriority`.  That's why  `-p 0` on the command line could be accepted, and translated into  `defaultPriority (99)`. 
=====)

boombatower:
Thank you for the prompt review and merge. I see you already followed up on the -p 0. :)

PR #82, opened 2015-11-03
Suggest setting a non-default repo priority.

boombatower:
Encourage discovery of repository priorities which are important for maintaining
a stable system in general and especially when using Tumbleweed as many users
dup regularly. The messages do not interfere with normal behavior, but instead
focus on discovery rather than annoying prompts.

Messages are displayed:
- after `zypper addrepo` when using default priority
- before `zypper dup` with all default priorities

This is similar to the various helpful suggestions `git` makes. For example:

```
$ git status
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        modified:   README.txt

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   benchmarks.rb
```

I tried to keep the wording as condensed as possible since these are just tips and the man pages should be consulted for more detail. From the man page:

```
           -p, --priority positive-integer
               Set priority of the repository. Priority of 1 is the highest, the higher the number the lower the priority. Default
               priority is 99. Packages from repositories with higher priority will be preferred even in case there is a higher
               installable version available in the repository with a lower priority.
```

Which seems sufficient. If desired I can followup by adding some example situations to the man pages to make it clearer.

### DIFF
--- a/src/Zypper.cc
+++ b/src/Zypper.cc
@@ -4816,11 +4816,35 @@ void Zypper::doCommand()
 
     if (!copts.count("repo") && !copts.count("from")
         && repoManager().knownRepositories().size() > 1)
+    {
       out().warning(str::form(_(
         "You are about to do a distribution upgrade with all enabled"
         " repositories. Make sure these repositories are compatible before you"
         " continue. See '%s' for more information about this command."),
         "man zypper"));
+
+      // check to see if all repos have the default priority
+      bool defaultPriority = true;
+      RepoInfoList repos = repoManager().knownRepositories();
+      for ( RepoInfoList::iterator it = repos.begin(); it != repos.end(); ++it )
+      {
+        RepoInfo & nrepo( *it );
+        if ( nrepo.priority() != RepoInfo::defaultPriority() )
+        {
+          defaultPriority = false;
+          break;
+        }
+      }
+
+      if ( defaultPriority )
+      {
+        out().warning(_(
+          "All repositories have the default priority."
+          " Use \"zypper modifyrepo --priority <integer> ...\" to set priorities,"
+          " lower numbers for repositories that replace default packages and higher"
+          " for missing packages."));
+      }
+    }
 
     init_target(*this);
     init_repos(*this);
--- a/src/repos.cc
+++ b/src/repos.cc
@@ -1755,6 +1755,13 @@ void add_repo(Zypper & zypper, RepoInfo & repo)
     // translators: property name; short; used like "Name: value"
     p.add( _("URI"),		repo.baseUrlsBegin(), repo.baseUrlsEnd() );
     s << p;
+
+    if ( repo.priority() == RepoInfo::defaultPriority() )
+    {
+      s << endl << "  " <<
+        _("(use \"zypper addrepo --priority <integer> ...\" to ensure repositories are utilized as desired,"
+          " lower numbers for repositories that replace default packages and higher for missing packages)");
+    }
   }
   zypper.out().info(s.str());
 


mlandres:
Reluctant - actually we do not encourage using priorities; vendor protection is usually easier to handle and gives better results.

Since `zypper-1.12.13` even Tumbleweed users, who frequently `dup`, can enable 'vendor protection'. Darix [blogged](https://discourse.nordisch.org/t/living-on-tumbleweed/379) about it.

boombatower:
Based on what I read in the blog post and the related bug it seems there are still a few things to be worked out before it is a fully baked solution, much less default. Also looks like some similar "suggest" style addition could be helpful in polishing the feature.

It still seems that when performing a normal `dup` priorities are extremely useful and it seems like there will still be time when a normal `dup` is the right choice. If you do not have some of the special cases mentioned in the blog post (where you want to defy priorities) it seems like priorities are still possibly a better solution since they allow more automatic flexibility for package additions/removals within available repositories.

With that in mind, perhaps the best solution is to add a condition on these messages to check for `allow-vendor-change`. If that is allowed then the messages still make sense, perhaps a note to take a look at the `--no-allow-vendor-change` flag and settings.

If agreeable, I'll make that change the conditional change and followup with a few more "suggest" type changes for both priorities and the settings approach. One way or another these solutions need to be more visible.

boombatower:
Some other thoughts.

Using the "simple dup" approach you still need to make all the vendor changes. Either manually per-package or per repositories (with possible switches afterwards) `zypper dup --from <repo>`. Unless the users has a rather specific setup they will end up with packages from repositories based on priorities, but without the self-documentation and automatic vendor changes made available by just setting repo priorities. Instead when looking down the road at a package available from multiple repositories provides no indication of why.

Using priorities I can add multiple repos and dup in one step to packages provided by both, without fuss. If repos provide sensible defaults in `.repo` file (as an issue I filed against obs) then this becomes transparent. Use wants KDE stable updates? Add repo and dup.

If however one does have a rather specific "woven" case then manually picking and telling zypper to not change it makes a lot of sense.

boombatower:
bump, I see people in IRC everyday where this is the solution

sysrich:
I agree with @mlandres , we normally discourage the use of priorities.

Adding additional repositories should always be considered an 'advanced' option. By advanced I mean something along the lines of "the users doing it should be comfortable playing with stuff like priorities"

Even if we accepted PR's like this or your other to OBS, we'd just be encouraging bad practice, where users will be adding additional complexity to their systems that WILL break sooner or later, even with (in some cases especially with) repo priorities being set

For Tumbleweed the better long term solution is to encourage more repo maintainers to maintain their stuff in Tumbleweed, which eliminates the need for users to play with additional repositories and provides a nice consistent, built together, tested together, platform for use

boombatower:
People will be using packman (extremely common) for the foreseeable future and repos for things that maintainers haven't decided to put into factory or are rejected. The feature exists, someone felt it was important enough to build in the first place, but somehow it's a bad idea to have sensible defaults? Even the most advanced users like sensible defaults. One less thing I have to worry about.

Not only does zypper have repo priorities, but it even will read them form the .repo file, but it's a bad idea to allow developers and maintainers to set a sensible default? The fact that this feature is so hidden and not evangelized has led to lots of issues for all sorts of users seen in IRC and forum posts.

People _will_ break their system adding repos the standard way that everyone knows about (ie without priorities). They have hope with priorities. Encouraging different behavior from maintainers is great, but I can encourage people not to steal that doesn't mean I get rid of the police ahead of time.

Encouraging the ymp (single click install) will allow novice users to break their system even easier since they aren't even aware they're adding a repo, much less what that is. Having the project maintainer set a priority helps ensure there is a chance things will go well.

Just look at packman the most common non-base repo. Your media codecs will break almost every other update depending on the repo state and the flip flop one gets from zypper dup.

If the encouragement of putting things in Factory works, then having proper priorities means people will automatically switch to those packages when they are available...without even having to be aware of it.

Most "advanced' users testing out updated builds and bugfixes don't even know about repo priorities since it is such a well guarded secret. Until we live in the magical world with unicorns can we please get the tools to work in the present?

mlandres:
I don't mind raising the awareness of priorities by mentioning them more prominently in zypper command summaries, stats or the man page.

But zypper messages will have to be neutral. We also serve products which disfavor using priorities so zypper must not suggest using them or even warn if they are absent. But we may of course show that they do exist and explain how they work.

If you happen to have a few sentences explaining repo priorities for the man pages 'CONCEPTS/Repositories' or 'Repository Management' section, I will gladly use them.

I will look into it next week.

boombatower:
Fair enough.

Any thoughts about the suggestion showing up if using repos in addition to standard base with default priorities? Possibly also add a config option or compile flag so it can be disabled in SLE builds or what not if desired?

sysrich:
I do not know of any product using zypper that favours using priorities

They are not healthy

The problem you cite with packman is just as easily avoided by using zypper up or zypper dup --no-allow-vendor-change

I'd much rather see we lean on the established and healthy vendor mechanism rather than encouraging setting priorities for untreatable repositories

mlandres:
Zypper does already support priorities and the feature itself is neither good nor bad. I'm also using Tumbleweed + packman + various OBS repos and I think that my life is easier without priorities, but it's freedom of choice.

The IMO valid point is that zypper is somewhat neglecting the priorities. We neither help to maintain them nor to verify they are not in effect. Which is the better strategy will not be decided or suggested by zypper.

boombatower:
I'll have to read up on the vendor mechanism, but something should be done as people hose their systems repeatedly in IRC since the default, in all wiki pages, leads to easy breakage.

mlandres:
For zypper-1.13.13 I added a priority summary to the add-repo and list-repo commands. It will tell you if:
```
Repository priorities are without effect. All enabled repositories share the same priority.
```
or something like this:
```
Repository priorities in effect:                          (See 'zypper lr -P' for details)
      20 (raised priority)  :  1 repository  
      99 (default priority) : 12 repositories
```
A section for the man page will follow...

boombatower:
I noticed that the other day and was pleasantly surprised. Very nice.

boombatower:
@sysrich: After thinking about what you said some more it does not make sense to use trust as a reason against priorities. Any package can provide a systemd service (or similar) and enable it to start on boot and do whatever they want. I can patch any software package in a way that is malicious. Having the ability to override "core" packages is completely unnecessary to perform all sorts of nasty things. There is implicit trust when you add a repo to your system. There is also clearly a reason for adding a repo which is communicated by the priority.

Can you elaborate on what you are referring to by the "vendor mechanism" or point me to some documentation?

The lack of a solution by default one way or another hit another user today in IRC (daily occurrence):
```
[15:42] <kacper> ok it works, after I completly switched to packman
```

mlandres:
https://doc.opensuse.org/projects/libzypp/HEAD/pg_zypp-solv-vendorchange.html

    In short: When looking for an installed packages update candidate, 
    we are  looking for a package originated by the same vendor (not repository!) 
    as the installed one.


The idea behind it is that you use `zypper dup` (dist-upgrade) to move from one distro version to another. This is the only command that has _vendor protection disabled_ (also allows arch change and downgrades) because we arrange all packages the same way as if you had installed the new distro repos.

After the dist-upgrade you maintain your system with `up` or `patch`. These commands prefer to stay with the same vendor and architecture as an already installed package and don't downgrade. So they try to keep things as you have arranged them.

TW however is different, as people tend to use `dup` to maintain it. By this they lose the vendor protection and their packman packages get overwritten by suse ones and vice versa. That's why we introduced `dup  --no-allow-vendor-change` (still allows arch change and downgrades). As TW users may want this being the default for their `dup` it can be set in `/etc/zypp/zypp.conf':

```
## EXPERTS ONLY: TUNE DISTRIBUTION UPGRADE (DUP)
## Set whether to allow changing the packages vendor upon DUP. If you
## are following a continuous distribution like Tumbleweed or Factory
## where you use 'zypper dup --no-allow-vendor-change' quite frequently,
## you may indeed benefit from disabling the VendorChange. Packages from
## OBS repos will then be kept rather than being overwritten by Tumbleweeds
## version.
##
## Valid values:  boolean
## Default value: true
##
solver.dupAllowVendorChange = false
```

I prefer the vendor approach, because here my systems content defines the update (packman to packman, obs to obs, suse to suse,...).  Packages stay as I have arranged them. If I want to switch packages to packman, I do `zypper dup --from packman`. I can switch individual packages back to suse or some other vendor, and they will stay with that source. No matter what the other repo vendors do.

With repo priorities the repo content defines the update. Here you rely on the ecosystem that surrounds you. Priorities and repo content define how your system looks like. It's a more passive approach, but if you trust the ecosystem and it is healthy - fine. But as zypper can't judge the ecosystem, zypper can't recommend using priorities. This has to be a decision of the individual running the system. 

Talking about security I don't see that much difference in both approaches. If I use a repository, I need to trust it's vendor. If a trusted vendor attacks or is compromised by an attacker, I am vulnerable. If I trust every vendor without asking, I'm easy to exploit.

mlandres:
Closing the issue. Changes in `zypper-1.13.13` raise the awareness of priorities being present, but without suggesting to change anything.

PR #84, opened 2015-11-05
Fix #83 incomplete bash completion for --type

### LINKED ISSUE #83: bash completion for --type incomplete
zypper manpage sais the following are avaialble for --type:
package, patch, pattern, product, srcpackage, application

Bash completion for zypper search --type only does package, patch, pattern, product.

andreasstieger:
Fixes #83


mlandres:
thanks

PR #103, opened 2016-10-23
bug fixes and translations

skriesch:
We can't use Weblate for translations in zypper at the moment. I want to add translations and translation bug fixes for the German team.


mlandres:
Danke.

PR #139, opened 2017-10-26
Trim excess nesting in addUpdateHint

jengelh:


mlandres:
Thanks, but IMO the nesting is OK as we may want to add some more cases here. Nested it's easier to extend.

jengelh:
Well you don't really know that until you need it. Just trying to do some awareness here before it ends in tears like bs_sched.

PR #141, opened 2017-11-17
Add Debian apt compat

bmwiedemann:
newer Debian versions use 'apt' as their preferred package management
tool - using mostly the same CLI as previous apt-get and aptitude.

Note: not tested


mlandres:
@bmwiedemann Sorry Bernhard, I missed this. Shall I merge it, or is it too untested?

bmwiedemann:
IMHO you can merge it. There is not much that can go wrong there.

PR #148, opened 2018-02-09
fix "it's" typo

aspiers:
"it's" is always short for "it is".  Here the word is in the possessive form, so there is no apostrophe.

https://en.oxforddictionaries.com/usage/its-or-it-s


mlandres:
Thanks.

PR #150, opened 2018-02-22
Enable debug file output for zypper ps

bzeller:
Needs to land with https://github.com/openSUSE/libzypp/pull/91


src/Zypper.cc
```diff
@@ -3200,6 +3201,8 @@ void Zypper::processCommandOptions()
 	     _("Create a short table not showing the deleted files. Given twice, show only processes which are associated with a system service. Given three times, list the associated system service names only.") )
     .option( "--print <format>",	// translators: --print <format>
 	     _("For each associated system service print <format> on the standard output, followed by a newline. Any '%s' directive in <format> is replaced by the system service name.") )
+    .option("-d, --debugFile <path>",
+       _("Write debug output to file <path>."))
```
mlandres: Precede translatable option description with `// translators: -d, --debugFile <path>`, like in the lines above.

Comments preceding a translatable string go into the .pot file. Hints for the translator are tagged as `// translators: `

doc/CMakeLists.txt
```diff
@@ -1,44 +1,51 @@
...
+
+    FOREACH (MAN_IDX RANGE 1 8)
+        IF (NOT DEFINED A2X_MAN_PAGES${MAN_IDX})
+            CONTINUE()
+        ENDIF()
+
+        FOREACH (curr_PAGE ${A2X_MAN_PAGES${MAN_IDX}})
+            ADD_CUSTOM_COMMAND (
+              OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${curr_PAGE}
+              COMMAND ${A2X} -D ${CMAKE_CURRENT_BINARY_DIR} -d manpage -f manpage ${CMAKE_CURRENT_SOURCE_DIR}/${curr_PAGE}.txt
```
mlandres: As in https://github.com/openSUSE/libzypp/pull/91#discussion-diff-171534971R30

src/Zypper.cc
mlandres: Sorry, I forgot: Please update the man page accordingly.

PR #151, opened 2018-02-26
Deprecate --type switch in ar and as

bzeller:


src/Zypper.cc
```diff
@@ -340,22 +340,39 @@ namespace {
...
-			 : _("Legacy commandline option %1% detected. Please use %2% instead.") )
-			 % NEGATIVEString(dashdash(old_r))
-			 % POSITIVEString(dashdash(new_r));
+  enum LegacyCLIMsgType {
+    Local,
+    Global,
+    Ignored
+  };
+
+  inline std::string legacyCLIStr( const std::string & old_r, const std::string & new_r, LegacyCLIMsgType type )
```
mlandres: Throughout the reviews I realize how much I actually miss function args not following the `_r` convention. I'd vote for keeping this as a rule (for new code). So `type_r`.

Throughout my code there are just a few exceptions. Binary operator  arguments are usually just `lhs` and `rhs` and ostream operator<< uses `str` and `obj`. The rest usually follows the `_r` convention.

src/Zypper.cc
```diff
@@ -1918,7 +1935,8 @@ void Zypper::processCommandOptions()
     )
     .optionSectionCommandOptions()
     .option_SERVICE_PROP
-    .option( "-t, --type <TYPE>",	(str::Format(_("Type of the service (%1%).") ) % "RIS").str() )	// FIXME: leagcy, actually autodetected but check libzypp
+    .legacyOptionSection()
+    .option( "-t, --type <TYPE>",	(str::Format(_("Type of the service (%1%). This option is ignored.") ) % "RIS").str() )	// FIXME: leagcy, actually autodetected but check libzypp
```
mlandres: I wouldn't even repeat the old description, just `_("The type of service is always autodetected. This option is ignored.")

src/Zypper.cc
```diff
@@ -2131,7 +2149,8 @@ void Zypper::processCommandOptions()
     .option( "-C, --no-check",		_("Don't probe URI, probe later during refresh.") )
     .gap()
     .option_REPO_PROP
-    .option( "-t, --type <TYPE>",	str::Format(_("Type of repository (%1%).") ) % "yast2, rpm-md, plaindir" )	// FIXME: leagcy, actually autodetected but check libzypp
+    .legacyOptionSection()
+    .option( "-t, --type <TYPE>",	str::Format(_("Type of repository (%1%). This option is ignored.") ) % "yast2, rpm-md, plaindir" )	// FIXME: leagcy, actually autodetected but check libzypp
```
mlandres: Same as with the service description.

src/Zypper.cc

src/Zypper.cc

src/Zypper.cc

PR #158, opened 2018-03-08
Introduce --allow-unsigned-rpm switch to install (bsc#1055533)

bzeller:


src/Zypper.cc
```diff
@@ -4270,6 +4272,13 @@ void Zypper::doCommand()
       repo.setAutorefresh( true );
       repo.setKeepPackages( false );
 
+      if ( _copts.count("allow-unsigned-rpm") )
+      {
+        repo.setGpgCheck(RepoInfo::GpgCheck::AllowUnsignedPackage);
+      }
+
+
+
```
mlandres: 1 or 2 empty line less should do

src/Zypper.cc
```diff
@@ -1721,6 +1722,7 @@ void Zypper::processCommandOptions()
       "    --download              Set the download-install mode. Available modes:\n"
       "                            %s\n"
       "-d, --download-only         Only download the packages, do not install.\n"
+      "    --allow-unsigned-rpm    Silently install unsigned rpm packages given as commandline parameters.\n"
```
mlandres: That's one of the ugly cases, where adding/changing a CLI option breaks the translated monster-help-text. We try hard not to do this.

If it is acceptable to put the new option at the end of the help text, make sure the `_command_help =` assignment already uses the `( CommandHelpFormater() << _("the original translated help text maybe in a str::Format") )`. Then new options and option sections can be added using the Fomater methods. One is already present at the install help:

    .option( "-y, --no-confirm",	_("Don't require user interaction. Alias for the --non-interactive global option.") )

You can add yours in front or below.


If adding the option at the end is too bad (or I have very much time), I split the original help text into pieces which can be used with the CommandHelpFormater method. So I can rearrange the help text as needed. In order not to lose the translations, .po files containing translation of the help text must be adjusted accordingly. That's no fun.

src/Zypper.cc

src/Zypper.cc

PR #162, opened 2018-03-13
Add --installroot flag (fixes #157)

### LINKED ISSUE #157: feature request: --installroot option
dnf ships an `--installroot` option, which the [man page](https://dnf.readthedocs.io/en/latest/command_ref.html) describes as follows:

> Specifies an alternative installroot, relative to where all packages will be installed. Think of this like doing chroot <root> dnf except using --installroot allows dnf to work before the chroot is created.

Zypper's `--root` option comes close to that, but it does slightly more than necessary as `--installroot` will still use the hosts repositories, caches, metadata, etc. I have a specific use-case for such a feature in zypper, as I'd love to mount a container's rootfs (with no zypper installed to reduce size) and use the host's zypper to, for instance, install a new package in the container's rootfs.

bzeller:


src/repos.cc
```diff
@@ -999,7 +999,10 @@ void init_target( Zypper & zypper )
 
     try
     {
-      God->initializeTarget( zypper.globalOpts().root_dir );
+      if (zypper.globalOpts().install_root_dir.empty())
+        God->initializeTarget( zypper.globalOpts().root_dir );
+      else
+        God->initializeTarget( zypper.globalOpts().install_root_dir );
```
mlandres: I would not do it this way (i.e. just re-rooting the Target to the install_root_dir).

If you follow `--root` (usage of `root_dir`), you will find that several code locations refer to it when retrieving data (e.g. locks, product data,..). Re-rooting the Target alone is not sufficient for a correct workflow.

I'd start behaving exactly as if `--root` was passed (set root_dir, changedRoot,...). Just the `RepoManagerOptions` should be rooted to use `"/"` rather than `_gopts.root_dir`. 

This way you behave like `--root`, just the repos are taken from the outer system (this is also how we should describe it in the help and man page). Due to the repo refresh `--installroot` should require being `root`, which is not enforced with `--root`.

BTW both are mutual exclusive; there should be an error if --root and --installroot are specified at the same time..

bzeller: Done

bzeller: Done except the superuser part, do we need to enforce that? For example a zypper se inside the installroot would not require superuser privs?

src/Zypper.cc
```diff
@@ -1220,10 +1222,17 @@ void Zypper::processGlobalOptions()
...
   }
 
-  if ( (it = gopts.find("root")) != gopts.end() )
+  if ( (it = gopts.find("root")) != gopts.end() || (it = gopts.find("installroot")) != gopts.end() )
   {
+    if ( gopts.find("root") != gopts.end() && gopts.find("installroot") != gopts.end() )
+    {
+      out().error(_("--root can not be used together with the --installroot flag."));
+      setExitCode( ZYPPER_EXIT_ERR_INVALID_ARGS );
+      return;
```
mlandres: I know its 'Kraut und Rüben' in the code. The desired style until the error handling is redesigned is throwing an ExitRequestException:
```
out().error(_("...") );
setExitCode( ZYPPER_EXIT_ERR_INVALID_ARGS );
ZYPP_THROW( ExitRequestException("invalid args") );
```

Two rules for the translated error messages as we (want to) avoid duplication:
 - Fix strings like option names should be injected via Format. I.e use one translated string for multiple cases.
 - If the same translated string is used multiple times and at multiple places, we should try to use a function.  providing/building the string. I recently created a 'cli' namespace in Zypper.cc to collect all such stuff. If you rebase on 67ad041420186ca54597dbffb050905efa5dc875 you can use this message:

```
  out().error( cli::errorMutuallyExclusiveOptions( "--root --installroot" ) );
  ...
```

src/Zypper.cc
```diff
@@ -698,6 +698,7 @@ void print_main_help( Zypper & zypper )
 
   static std::string help_global_target_options = _("     Target Options:\n"
     "\t--root, -R <dir>\tOperate on a different root directory.\n"
+    "\t--installroot <dir>\tOperate on a different root directory, but share repositories with the host.\n"
     "\t--disable-system-resolvables\n"
     "\t\t\t\tDo not read installed packages.\n"
```
mlandres: Like with the command options, we should not change the existing translated string. For now I'd append the option:
```
help_global_target_options += str::Format( "\t--installroot <dir>\t%1%\n" ) % _(Operate on a different root directory, but share repositories with the host.)
```

bzeller: Sorry i keep forgetting that

src/Zypper.cc

bzeller:
Do we really need to enforce that root priviledges? For example a zypper se inside the installroot would not require superuser privs?

mlandres:
> Do we really need to enforce that root priviledges?

Well, `zypper se` is a (more or less) readonly operation. If a repo is outdated and can not be refreshed, you should see a note and know the search result may be inaccurate:

    $ zypper se xxx
    Repository 'CloudRepo' is out-of-date. You can run 'zypper refresh' as root to update it.
    Repository 'OpenStack (42.3)' is out-of-date. You can run 'zypper refresh' as root to update it.
    ...

IMO we can stay with the `--root` behavior (non-root ok) if we are sure the above notes are shown if repos are outdated.

PR #163, opened 2018-03-14
[WIP] Add zypper keys subcommand

bzeller:


mlandres:
A good start. 2 things immediately caught my eye:

1. The commands probably don't need to launch the repos (refresh and loading the repo data). Initializing the Target should be enough.
2. (nit-picky) The term `subcommand` is used for external programs just hooked into zypper (`zypper help subcommand`). `zypper keys` will become a (builtin) `command` :)

src/keys.cc
```diff
@@ -0,0 +1,223 @@
...
+  else
+  {
+    const target::rpm::RpmDb &rpmdb = God->target()->rpmDb();
+    dumpKeyList( zypp_r, rpmdb.pubkeys(), zypp_r.cOpts().count("details") );
+  }
+}
+
+void removeKey ( Zypper &zypp_r, const std::string &id )
+{
+  if ( !God || !God->keyRing() )
```
bzeller: need to request root privs here

src/keys.cc
```diff
@@ -0,0 +1,223 @@
...
+  {
+    zypp_r.out().warning( str::Format(_("Key %1% is not known")) % id );
+    zypp_r.setExitCode( ZYPPER_EXIT_ERR_INVALID_ARGS );
+    return;
+  }
+}
+
+void importKey ( Zypper &zypp_r, const std::string &url_r )
+{
+  if ( !God || !God->keyRing() )
```
bzeller: need to request root privs here

src/keys.cc
bzeller: Done

src/keys.cc
bzeller: Done

src/callbacks/keyring.h
```diff
@@ -50,7 +51,6 @@ namespace zypp
     {
       Zypper::instance().out().warningPar( 4, str::Format(_("This file was modified after it has been signed. This may have been a malicious change, so it might not be trustworthy anymore! You should not continue unless you know it's safe.") ) );
     }
-
     std::ostream & dumpKeyInfo( std::ostream & str, const PublicKeyData & key, const KeyContext & context = KeyContext() )
```
mlandres: Remove here if it's now in utils/misc.cc.

src/Zypper.cc
```diff
@@ -900,6 +924,12 @@ void print_main_help( Zypper & zypper )
     "\tzypper <subcommand> [--command-options] [arguments]\n"
   );
 
+  static std::string help_keys_commands = _("     Key Management:\n"
+    "\tkeys, lk\t\tList all keys.\n"
+    "\taddkey, as\t\tAdd a new key to trust.\n"
+    "\tremovekey, rs\tRemove a key from trust.\n"
```
mlandres: Short names `as/rs` are wrong (copied from asdd/remove service?). Must be `ak/rk`

src/Zypper.cc
```diff
@@ -5745,6 +5835,41 @@ void Zypper::doCommand()
...
+    // load system data...
+    init_target( *this );
+    if ( exitCode() != ZYPPER_EXIT_OK )
+      return;
+
+    importKey( *this, arguments().front() );
+    break;
+  }
+
+  case ZypperCommand::REMOVEKEY_e:
```
mlandres:
```
$ zypper rk
Segmentation fault (core dumped)
```

src/Zypper.cc
```diff
@@ -5745,6 +5835,41 @@ void Zypper::doCommand()
     break;
   }
 
+  case ZypperCommand::KEYS_e:
```
mlandres:
```
$ zypper keys 39DB7C82
+-------------------+
|   . . ^           |  Key ID:           70AF9E8139DB7C82
|    : : .          |  Key Name:         SuSE Package Signing Key <build@suse.de>
...
$ zypper keys 9DB7C82
Warning: Argument '9DB7C82' is not a trusted ID nor a existing file.
$ zypper keys @suse.de
Warning: Argument '@suse.de' is not a trusted ID nor a existing file.
+----[39DB7C82]-----+
```
What do you think about relaxing the required argument format? 

Maybe is 'trailing portion of an `ID`' or 'is substring of the `Name`' (all case insensitive).

Same could be done for `removekey`, with the additional constraint, that every arg must match exactly 1 key.  If an arg matches more one could say `"ARG" specifies multiple  keys' (opt. indented list of maches 'ID Name').
`Use --all-matches to remove all metchein keys`.

src/Zypper.cc

src/callbacks/keyring.h
bzeller: Done

src/Zypper.cc

src/Zypper.cc

Issue #89, opened 2015-12-15
Support for exposing /var/lib/zypp/AutoInstalled


alexjp:
Hi,

Support for exposing /var/lib/zypp/AutoInstalled functionality for marking packages as AutoInstalled or Explicit.

I was thinking of something like:
```
zypper se --autoinstalled ( list packages that were autoinstalled and present on /var/lib/zypp/AutoInstalled )
zypper se --explicit ( list packages installed on system but no on autoinstalled file )

zypper mark --autoinstalled "packages" ( mark, add packages to autoinstalled file )
zypper mark --explicit "packages" ( mark, remove packages from autoinstalled file )

zypper rm --unneeded ( remove packages not needed on autoinstalled file ).
```

I have been using these scripts to manage autoinstalled packages: https://github.com/alexjp/zypper_autopackages

 zypper_autoinstalled.sh: creates a full autoinstalled file from packages not present in a world file
 zypper_autoremove.sh: remove all packages not needed by packages not in autoinstalled
 zypper_explicit_installed.sh: creates the world file: all packages installed but no in autoinstalled

With these scripts, I can have a common world file with favorite packages, and manage those across diferent machines.
With zypper_autoremove.sh, I can have the peace of mind that i can install all i want, if later is not in the world file, it can be cleaned and removed from system.  ( very helpfull when installing and testing a new desktop environment or group of packages ).

Thanks!

mlandres:
Indeed a nice feature.

mlandres:
The idea of a `world file with favorite packages` is a bit related to `/etc/zypp/systemCheck`.

The idea was to be able to provide a black/white-list of packages required/forbidden on a system. Each and every solver run has the systemCheck rules enabled. It's the reason why you can hardly `zypper rm glibc`. Custom settings can be defined as file in `/etc/zypp/systemCheck.d`.

alexjp:
Yes, i guess i was doing my own custom systemCheck.d. I didnt knew about systemCheck file.
The systemCheck.d folder is a nice touch ( example for kde packages, gnome packages, sys packages ).

Seems opensuse and zypper has all the infrastructure to allow this features. Nice!

ykurlaev:
Is PR #165 related to this issue? Is there any work being done on it currently?

Xplosm:
> Is PR #165 related to this issue? Is there any work being done on it currently?

Yes, I believe so. And no. It seems no work has been done in either PR.

PR #165, opened 2018-03-29
Add UI to manage the auto install flag on packages

bzeller:
Adds a zypper command called "mark" that gives the user the possiblity to set and remove certain flags on installed packages. Currently only manual and auto can be set. 

zypper mark auto packge1 package2


bmwiedemann:
This feature seems useful. What is left to do?

martinjungblut:
@mlandres Sorry if I'm bumping this without a greater understanding, but what is left for this code to be merged? It'd be really useful.

Xplosm:
I believe this PR is related to [this one](https://github.com/openSUSE/zypper/pull/89).

How is this evolving?

francoism90:
I would like to have this, is this still not supported? I cannot seem to find any command that does this for me.

bmwiedemann:
At least the data is available in plain text form, so that it could be edited manually with `sudo vim /var/lib/zypp/AutoInstalled`

In Debian, there is `apt markauto $PKG` and `apt unmarkauto $PKG` and it should not be hard to add an equivalent `zypper mark manual $PKG`

PR #173, opened 2018-05-18
Support listing gpgkey URLs in repo files (bsc#1088037)

bzeller:
Handles the new report introduced in libzypp for accepting package signing keys received via gpgkey urls in repo configuration files.


Conan-Kudo:
Wait, this didn't work before?

bzeller:
No, those lines were ignored by libzypp and zypper before

Conan-Kudo:
@bzeller You know that can be a newline or whitespace separate list, just like in `baseurl`, right?

bzeller:
I did not touch the parser code for that. Libzypp always did parse and store it inside RepoInfo but never made use of the parsed information. If that parser is not handling all cases correctly that would be another feature/bug to look at.

PR #175, opened 2018-05-23
Add consistent repo selection options to rr

bzeller:
Provide the same set of aggregate options for "removerepo" as for "modifyrepo" in zypper.


bzeller:
Updated the man page

PR #179, opened 2018-06-08
Support asking the user about rpm header downloads

bzeller:
Makes use of the new PackageHeadReportReceiver to ask the user if a RPM header should be downloaded. Also adds possibility to query changelog and filelists.


bzeller:
Fixes bug: bnc#485945

PR #181, opened 2018-06-14
Add setter functions for zypp cache related config values to ZConfig (fixes #180)

### LINKED ISSUE #180: zypper creates file /var/cache/zypp/solve despite --solve-cache-dir being used
We are currently working an issue in `zypper-docker` where we want to avoid changing the potentially used zypper cache by letting zypper point to a temporary directory that can be used as an intermediate cache that will be removed afterwards.

@parlt91 is working on adding integration tests and noticed that zypper still creates files in /var/cache/zypp/solve despite being called with `-C` or `--solve-cache-dir`.

A reproducer can look as follows (run in a tumbleweed containers):
- `$ zypper ref`
- `$ find /var/cache/zypp -type f -exec md5sum {} \; > initial.cache`
- `$ mkdir /foo`
- `$ zypper -C /foo in tar`
- `$ find /var/cache/zypp -type f -exec md5sum {} \; > after-install.cache`

When diffing the `initial.cache` and `after-install.cache` we see that the following files changed:
``` diff
@@ -1,6 +1,6 @@
-408336acbf81cb7dd697a67ebfc21bfc  /var/cache/zypp/solv/@System/cookie      
-1f009f8ad204c0c0f48a5ddf00cfe7a5  /var/cache/zypp/solv/@System/solv        
-899e09186f33112a76f5f17864ad8f6c  /var/cache/zypp/solv/@System/solv.idx    
+772a9ef2386c767a3a94829ba0a656f4  /var/cache/zypp/solv/@System/cookie      
+1dbd69477ee28dd66536fff7d5e301a8  /var/cache/zypp/solv/@System/solv        
+2a7813afb9f22875317e8abf64ae638d  /var/cache/zypp/solv/@System/solv.idx    
 9065281067eac3446123face1dc048c6  /var/cache/zypp/solv/NON OSS/solv        
 dfc28ff223632cf4f942f61c5ceff389  /var/cache/zypp/solv/NON OSS/solv.idx    
 836e4a5293074d398a4420dc11560b4f  /var/cache/zypp/solv/NON OSS/cookie      
```

I am not sure if this behavior is intended, but it is certainly surprising as the manpage suggests differently.

Cc: @flavio, @mssola

bzeller:
Fixes issue #180 by overriding the ZConfig zypp cache values based on the switches that are passed to zypper on cli. Currently those are set only in repomanager options.


src/Zypper.cc
```diff
@@ -1362,20 +1362,27 @@ void Zypper::processGlobalOptions()
 
   if ( (it = gopts.find("cache-dir")) != gopts.end() )
   {
+    ZConfig::instance().setRepoCachePath( it->second.front() );
     _gopts.rm_options.repoCachePath		= it->second.front();
     _gopts.rm_options.repoRawCachePath		= _gopts.rm_options.repoCachePath / "raw";
     _gopts.rm_options.repoSolvCachePath		= _gopts.rm_options.repoCachePath / "solv";
     _gopts.rm_options.repoPackagesCachePath	= _gopts.rm_options.repoCachePath / "packages";
   }
```
mlandres: In fact zypper should also not know about `"raw"`, `"solv"` and `"packages"`.  IMO ugly enough that zypper needs to know which dependent paths need to be reset, after `repoCachePath` has canged.
Shouldn't we assign `ZConfig::instance().repoMetadataPath()`, `repoSolvfilesPath()` and `repoPackagesPath()` instead?

bzeller: Right, then the logic for that is all in ZConfig

mlandres:
Some issue that's now harder to solve with PRs: This PR depends on openSUSE/libzypp#117. I.e. zypper `BuildRequires` at least the libzypp version which has the libzypp PR merged.

If we put it the other way: After the libzypp PR was merged, anticipate the next libzypp version (current patchlevel++), and inject a single commit like611a0fca9f69939ebca60b61c470b5fa6e531222 (ideally before the commit that requires the libzypp change). To avoid merge conflicts this should be done in the PR immediately before merging, or directly in git.

bzeller:
Updated the BuildDepends

PR #183, opened 2018-06-27
Out of bounds read in mbs iterator

bzeller:
Fix one possible and one definitive out of bounds read in MbsIterator:

Out of bounds read happens in that case:
-> previous read reads last char 
-> current read advances pos and sets _trest to 0 
-> mbrtowc will return error -2 ( because max read bytes is 0 ? )
-> we enter if ( _tread >= (size_t)-2 ) 
-> sets _tread to 1 
-> while loop will continue behind the last char with _tpos+_tread

Also added a testcase that triggers that behaviour by setting up a teststring that has 
multibyte continuation characters after a \0.

### DIFF
--- a/src/utils/text.h
+++ b/src/utils/text.h
@@ -122,9 +122,17 @@ namespace mbs
     {
       if ( !atEnd() )
       {
-	_tpos += _tread;
-	_trest -= _tread;
+        _tpos += _tread;
+        _trest -= _tread;
+
+        //we hit the end
+        if ( _trest == 0 ) {
+          setToEnd();
+          return *this;
+        }
+
 	_tread = ::mbrtowc( &_wc, _tpos, _trest, &_mbstate );
+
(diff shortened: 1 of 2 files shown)


src/utils/text.h
```diff
@@ -120,11 +120,24 @@ namespace mbs
 
     MbsIterator & operator++()
     {
+      auto setToEnd = [&](){
```
mlandres: Why a lamda and not a private inlined method?

bzeller: Habits, inlined code can cause trouble especially for libraries. E.g. if you have code in header the symbols are not exported. So when you change it and applications using your library are not recompiled when a new version was released you end up with 2 versions of the function. One in the binary and one in the library.

bzeller: Which of course it irrelevant in code that is already in a header ;)

PR #188, opened 2018-07-17
Sort search results by multiple columns (bsc#1066215)

bzeller:
This patch allows us to sort a table by multiple columns and a per row user value


src/Zypper.cc
```diff
@@ -4831,16 +4831,16 @@ void Zypper::doCommand()
         if ( command() == ZypperCommand::RUG_PATCH_SEARCH )
         {
           if ( copts.count("sort-by-repo") )
-            t.sort( 1 );
+            t.sort( { 1, Table::UserData } );
```
bzeller: zypper pse actually does not accept --sort-by-repo, dead code?

PR #191, opened 2018-07-23
Add expert (allow-*) options to zypper in (bsc#428822)

bzeller:
This patch adds the same set of expert switches to all zypper commands that could install packages as they are available for zypper dup:

--allow-downgrade
--no-allow-downgrade
--allow-name-change
--no-allow-name-change
--allow-arch-change
--no-allow-arch-change
--allow-vendor-change
--no-allow-vendor-change

### DIFF
--- a/doc/zypper.8.d/option_Solver_Flags_Installs.txt
+++ b/doc/zypper.8.d/option_Solver_Flags_Installs.txt
@@ -1,0 +1,11 @@
+	*--*[*no-*]*allow-downgrade*::
+		Whether to allow downgrading installed resolvables.
+
+	*--*[*no-*]*allow-name-change*::
+		Whether to allow changing the names of installed resolvables. Setting this to *no* will not replace packages which have been renamed.
+
+	*--*[*no-*]*allow-arch-change*::
+		Whether to allow changing the architecture of installed resolvables.
+
+	*--*[*no-*]*allow-vendor-change*::
+		Whether to allow changing the vendor of installed resolvables. Setting this to *no* might be useful if you do not want packages from foreign repos being changed to the distributions version (or vice versa).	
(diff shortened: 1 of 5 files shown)


mlandres:
As always:  if for `install`, then what about the other commands? 

We have  `install | remove | update | dup | patch | verify | inr`.  `remove` is special as it can not install packages (while install can remove).. `inr` (currently) because it requires recommends being turned on. 

But basically we have [`ARG_Solver_Flags_Common`/`option_Solver_Flags_Common`](https://github.com/openSUSE/zypper/blob/master/src/Zypper.cc#L152) for solver options which apply to all inst/remove commands.

A similar option set e.g. `ARG_Solver_Flags_Installs`/`option_Solver_Flags_Installs` which applies to everything except `remove` is not yet defined. This would be similar to [`ARG_Solver_Flags_Recommends	`/`option_Solver_Flags_Recommends`](https://github.com/openSUSE/zypper/blob/master/src/Zypper.cc#L164), but also applying to `inr`.

(The counterpart, options which do apply to `remove` only, do not yet exist.)


I'd create such an options definition for the _installs, define the [no-]allow-... options there and install  it for every command that has `ARG_Solver_Flags_Common`, except `remove`.

src/RequestFeedback.cc
```diff
@@ -117,9 +117,8 @@ std::string SolverRequester::Feedback::asUserString( const SolverRequester::Opti
 
     case UPD_CANDIDATE_CHANGES_VENDOR:
     {
-      PoolItem highest = ui::asSelectable()(_objinst)->highestAvailableVersionObj();
       std::ostringstream cmdhint;
-      cmdhint << "zypper install " << highest.asString();
+      cmdhint << "zypper install --allow-vendor-change " << _objinst.name();
```
mlandres: I would not change the hint here. `"zypper install " << highest.asString()` gives you exactly the reported package. No opt for the solver to change it.

bzeller: Done

PR #201, opened 2018-08-24
Detect read only filesystem on system modifying operations (fixes #199)

### LINKED ISSUE #199: Usability on transactional server
[[Feature request] Implement transactional-update into zypper](https://github.com/openSUSE/zypper/issues/198) made me think about the usability of zypper on a transactional server. Using zypper on a transactional server directly is pointless, because the root-fs is read-only. That is perfectly clear. But zypper still lets you happily type `zypper dup` and then starts to fetch 300 new packages. Once the installation starts and you'll get something like this:

```
Checking for file conflicts: ..................................................................................[done]
(  1/374) Removing libx86emu1-1.14-1.2.x86_64 ................................................................[error]
Removal of (54167)libx86emu1-1.14-1.2.x86_64(@System) failed:
Error: Subprocess failed. Error: RPM failed: error: can't create transaction lock on /var/lib/rpm/.rpm.lock (Read-only file system)
```

There are three scenarios why this might have happen:

1. The admin is not aware of how to interact with a transactional server and is lost now.
2. The admin forgot that this is not a traditional server and tried to upgrade it in the usual way.
3. This shows up in the logs or error messages of SaltStack, because the admin used `pkg.update` to update the system.

The first case is bad, because now the admin has to search the internetz to find a solution. The second case is also bad, because the seasoned admin has wasted precious time (and bandwidth) only to see that she/he should have used `transactional-update`. The third case might also stem from case one or two, but is only much better hidden in an error message.

I see two possibilities here:

1. Let zypper fail immediately and show a message to the user.
2. Don't expose zypper on a transactional server.

Okay, maybe I should have filed the issue somewhere else, but let's just start the discussion here. :wink: 

/cc @sysrich

bzeller:
This patch auto detects if the target filesystem is read only and if it is a transactional server. Zypper will exit directly and print a error message.

### DIFF
--- a/src/Zypper.cc
+++ b/src/Zypper.cc
@@ -18,6 +18,9 @@
 
 #include <unistd.h>
 #include <readline/history.h>
+#include <sys/vfs.h>
+#include <sys/statvfs.h>
+#include <linux/magic.h>
 
 #include <zypp/ZYppFactory.h>
 #include <zypp/zypp_detail/ZYppReadOnlyHack.h>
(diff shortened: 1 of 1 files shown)


src/Zypper.cc
```diff
@@ -488,6 +491,58 @@ namespace {
...
+
+        int err = 0;
+        do {
+          err = statfs( gopts_r.root_dir.c_str(), &fsinfo );
+        } while ( err == -1 && errno == EINTR );
+
+        if ( !err ) {
+          if ( fsinfo.f_flags & ST_RDONLY ) {
+
+            bool isTransactionalServer = ( fsinfo.f_type == BTRFS_SUPER_MAGIC && PathInfo( "/usr/sbin/transactional-update" ).isFile() );
```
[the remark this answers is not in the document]
bzeller: good idea, but we have no test suite around for that, so it would be best if you could provide that via PR, just make sure to add a cmake switch to enable it for builds where it is relevant

src/Zypper.cc
```diff
@@ -488,6 +491,61 @@ namespace {
...
+
+            str::Str msg;
+            if ( isTransactionalServer && !gopts_r.changedRoot ) {
+              msg << str::Format(_("This is a transactional-server, please use transactional-update to update or modify the system."));
```
mlandres: No `str::Format` needed as there's no placeholder inside. Same in L533. So a `std::string msg` in L529 is sufficient.

src/Zypper.cc
```diff
@@ -488,6 +491,61 @@ namespace {
...
+            } else {
+              msg << str::Format(_("The target filesystem is mounted as read-only. Please make sure the target filesystem is writeable."));
+            }
+            zypper.out().errorPar( msg );
```
mlandres: Also `ERR << msg << endl;` so we see also it in the log.

src/Zypper.cc
```diff
@@ -488,6 +491,61 @@ namespace {
...
+            return false;
+          }
+        } else {
+          MIL << "Checking if " << gopts_r.root_dir << " is mounted read only failed with errno : " << errno << std::endl;
```
mlandres: I'd us a `WAR <<`

src/Zypper.cc

src/Zypper.cc

src/Zypper.cc

sysrich:
I think this PR might have been overkill

For example transactional-update doesn't have functionality for features like `modifyrepo`, `addrepo`, and the like.

Those NEED to be run using `zypper`, because they need to modify the (read-write overlayfs provided) `/etc` which is then read by transactional-update

Now this PR has been merged, it seems to be impossible to configure repositories on any transactional *SUSE system with a version of zypper containing this change

Such as demonstrated here: https://openqa.opensuse.org/tests/754175#step/first_boot/16

@laenion - PING - am I mis-reading the situation?

laenion:
No, I agree - we need zypper for repository management, i.e. all `ZypperCommand::*_REPO_e:` need to be removed from the list of excluded commands. I haven't been using the `service` commands yet, but this may be the same for them.

Generally speaking on a transactional system (at least) _/etc_ and _/var_ will be writable, so if a command (usually) only changes files in those directory it shouldn't be blocked.

There's one notable exception: When `refresh` does not only update the package list, but also has to import a new key, then zypper will fail (as the key would be stored in the RPM database on the read-only file system). This currently can't be helped, but is automatically corrected when performing a package operation with transactional-update _and having the corresponding option in the configuration file set_.

As I also found it very confusing having to use a different command on a transactional system my next major task is to start a POC for implementing transactional-update as a zypper plugin. There's no guarantee that this will work out of course, but the goal is to implement as much functionality as possible directly into zypper. In the end I hope we won't need this patch in a few months at all, so I'm even wondering if it would make sense to revert the patch completely...

bzeller:
Good to know, we will revisit the code in that case.

mlandres:
4003f697ee53313acbda499ff5ff027703f422ca allows repo commands on transactional-server [(bsc#1108999)](https://bugzilla.suse.com/show_bug.cgi?id=1108999). 
Released in `zypper-1.14.11`.

PR #203, opened 2018-09-05
New style ps command

bzeller:
This patch migrates the "ps" command to the new style commands.


bzeller:
Now part of the locks PR, not needed anymore

PR #204, opened 2018-09-06
Always set error status if any nr of unknown repositories are passed to lr and ref (BSC #1093103)

bzeller:
The previous fix did not cover the case when only a unknown repository was passed to zypper ref. I changed the code so that if there is any unknown repository passed error status will be set.

### DIFF
--- a/src/repos.cc
+++ b/src/repos.cc
@@ -1324,10 +1324,10 @@ void list_repos( Zypper & zypper )
   else
     print_repo_list( zypper, repos );
 
-  if ( repos.empty() )
+  if ( !not_found.empty() ) {
+      zypper.setExitCode( ZYPPER_EXIT_ERR_INVALID_ARGS );
+  } else if ( repos.empty() ) {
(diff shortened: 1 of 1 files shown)


src/repos.cc
```diff
@@ -1444,9 +1444,15 @@ void refresh_repos( Zypper & zypper )
...
+      zypper.out().error(_("Some of the repositories have not been refreshed because they were not known.") );
+      zypper.setExitCode( ZYPPER_EXIT_ERR_INVALID_ARGS );
+      return;
+  }
+  else if ( enabled_repo_count == 0 )
+  {
+    if ( !specified.empty() )
       zypper.out().warning(_("Specified repositories are not enabled or defined.") );
     else
       zypper.out().warning(_("There are no enabled repositories defined.") );
```
mlandres: This `else` also needs to set `zypper.setExitCode( ZYPPER_EXIT_NO_REPOS );`

PR #209, opened 2018-10-02
subcommand: optimise fd-closing when running a subcommand

cyphar:
Iterating over (2,RLIMIT_NOFILE] and attempting to close(2) each fd is
quite expensive for two reasons:

1. It is very unlikely that the process already has RLIMIT_NOFILE file
   descriptors open, so many of the close(2) operations were unnecessary.

2. Because this is done *immediately* before an exec it's more expensive
   than setting O_CLOEXEC (which is handled in-kernel during execve(2)
   and is thus likely cheaper).

On most machines RLIMIT_NOFILE is 1024, so the cost of this is not
noticeable. However, inside containers RLIMIT_NOFILE can be in excess of
1 million -- which results in 1000x more close(2) operations. RPM had a
similar bug in the past, and was fixed [using the same /proc/self/fd
walking trick][1].

[1]: https://github.com/rpm-software-management/rpm/commit/5e6f05cd8dad

Reported-by: Marco Vedovati <mvedovati@suse.de>
Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/src/subcommand.cc
+++ b/src/subcommand.cc
@@ -257,9 +257,29 @@ namespace	// command execution
     {
       //////////////////////////////////////////////////////////////////////
 
-      // close all filedesctiptors above stderr
-      for ( int i = ::getdtablesize() - 1; i > 2; --i )
-      { ::close( i ); }
+      // close all *open* file descriptors
+      std::list<std::string> fdlist;
+      int maxfd = STDERR_FILENO;
+      int ret = readdir( fdlist, "/proc/self/fd", /*dots*/false );
+      if ( ret != 0 )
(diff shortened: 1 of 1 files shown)


mlandres:
Thanks.

cyphar:
Whoa whoa -- wait a second! I noticed a mistake and was about to fix it!

cyphar:
Yes, it doesn't build. I already have fixed it though, I'll send a fixed patch in a second.

cyphar:
#210 should fix it (it builds on my machine now).

PR #210, opened 2018-10-02
subcommand: fix build failure

cyphar:
There was a trivial build failure in ca1a44bad47f ("subcommand: optimise
fd-closing when running a subcommand"). Fix that.

Fixes #209.
Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/src/subcommand.cc
+++ b/src/subcommand.cc
@@ -258,26 +258,26 @@ namespace	// command execution
       //////////////////////////////////////////////////////////////////////
 
       // close all *open* file descriptors
-      std::list<std::string> fdlist;
+      std::list<Pathname> fdlist;
       int maxfd = STDERR_FILENO;
       int ret = readdir( fdlist, "/proc/self/fd", /*dots*/false );
       if ( ret != 0 )
       {
         // cannot open /proc/self/fd, fall back to expensive close-all approach.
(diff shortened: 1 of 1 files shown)


src/subcommand.cc
```diff
@@ -258,27 +258,26 @@ namespace	// command execution
       //////////////////////////////////////////////////////////////////////
 
       // close all *open* file descriptors
-      std::list<std::string> fdlist;
```
mlandres: Why not just `std::list<Pathname> fdlist;` and `std::stoi(fdstr.c_str());` on L275

cyphar: Does that work? I'm more of a C developer and was trying to reuse `readdir` from `src/source-download.cc` but the errors made me switch to the C way of doing it. I'll give that a shot.

cyphar: Oh wow, that does work. I'll push that version -- thanks!

cyphar: Pushed.

cyphar:
Thanks! Sorry for breaking the build. :cold_sweat:

mlandres:
My fault - I should have enabled the CI verify it builds. So don't worry too much.

PR #216, opened 2018-10-30
Store raw arguments inside BaseCommand if required

bzeller:
This patch adds a way to retrieve the list of raw arguments and positional arguments at command execution time.


src/commands/basecommand.cc
```diff
@@ -85,6 +88,41 @@ std::string ZypperBaseCommand::description() const
   return _description;
 }
 
+int ZypperBaseCommand::parseArguments( Zypper &zypp, const int firstOpt )
```
mlandres: `Zypper &zypper[_r]` :)

PR #221, opened 2018-11-26
Ma fate#325599 search packages

mlandres:
[Fate#325599](https://fate.suse.com/325599): Prompt options to change the `zypper.conf` option to `always` or `never` are missing Until I get augeas to write a changed config file. By now you have to change it manually.
```
[search]

## Whether an available zypper-search-packages-plugin should be called at the end of each search command.
##
## Valid values: boolean incl. 'always' or 'never', or 'ask' (the default)
## Default value: ask
##
runSearchPackages = ask
```
Known issue: The `infoLRHint` function cuts the text if the terminal is not wide enough. That's a bug to fix.


mlandres:
With the last 2 commits the `'... in zypper.conf: '` messages should be readable on narrow terminals (wrapped rather than clipped).

src/commands/search/search-packages-hinthack.cc
```diff
@@ -6,50 +6,191 @@
...
+      private:
+	bool _wasBlocked;
+      } tmpUnblock;
+
+      librpmDb::db_const_iterator it;
+      return( it.findPackage( ZSPP_PackageName ) && it->tag_provides().count( ZSPP_CliFwdCap ) );
+    }
+
+    /** Print a hint about an available search-packages subcommand. */
+    void printSearchPackagesHint( Zypper & zypper_r, bool plgInstalled_r )
```
bzeller: Zypper & zypper_r -> didn't you say that convention is Zypper & zypper? I don't mind either but would be good to know before I clean up the  Zypper &zypp_r  in the commands PR

mlandres: I wanted to demonstrate the use of `_r` and overdid it :)

src/commands/search/search-packages-hinthack.cc
```diff
@@ -6,50 +6,191 @@
...
+      if ( ! out.typeNORMAL() || out.verbosity() < Out::NORMAL )
+	return false;	// xml output or quiet mode
+
+      if ( zypper_r.runningShell() )
+	return false;	// we don't support subcommands in shell
+
+      if ( zypper_r.globalOpts().non_interactive )
+	return false;	// Never forward in non-interactive mode (search-packages itself relies on this!)
+
+      if ( zypper_r.globalOpts().changedRoot )
```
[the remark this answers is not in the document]
mlandres: In general zypper global options are not permitted with subcommands (so no --root). PM would say 'if a user is experienced enough to type --root, he does not need the search-packages hack'. In fact I want to minimize the chance this disturbs any image building, etc. Apart from that I don't know if  `SUSEconnect` is able to run chroot.

src/commands/search/search-packages-hinthack.cc
mlandres: BTW in `Zypper &zypp_r ` I nagged about `zypp` (missing `er`), not so much about the `_r`.  The code uses both `zypper` and `zypper_r`.

PR #223, opened 2018-12-10
Allow more complex argument parsing in zypper

bzeller:
This patch introduces dependencies between flags, additionally each flag can have its own policy to force a even stricter parsing order.
Also adds the possibility to have pre and post hooks that are called before and after writing a flag.

### DIFF
--- a/src/utils/flags/flagtypes.h
+++ b/src/utils/flags/flagtypes.h
@@ -105,8 +105,11 @@ inline std::string argValueConvert ( const CommandOption &, const boost::optiona
   return *in;
 }
 
+template <>
+int argValueConvert ( const CommandOption &, const boost::optional<std::string> &in );
+
 template <template<typename ...> class Container, typename T >
-Value GenericContainerType  ( Container<T> &target_r, std::string hint, const std::string sep = "" ) {
+Value GenericContainerType  ( Container<T> &target_r, std::string hint = std::string(), const std::string sep = "" ) {
   return Value (
     noDefaultValue,
     [ &target_r, sep ] ( const CommandOption &opt, const boost::optional<std::string> &in ) {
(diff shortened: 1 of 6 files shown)


src/utils/flags/zyppflags.h
```diff
@@ -170,7 +191,7 @@ namespace ZyppFlags {
    * \returns The first index in argv that was not parsed
    * \throws ZyppFlagsException or any subtypes of it
    */
-  int parseCLI ( const int argc, char * const *argv, const std::vector<CommandGroup> &options, const int firstOpt = 1 );
+  int parseCLI (const int argc, char * const *argv, const std::vector<CommandGroup> &options, const int firstOpt = 1 );
```
mlandres: I don't see the rule for it ;)

In the 2nd hunk src/utils/flags/flagtypes.cc e.g. you add the missing WS before the closing ')' (to make it consistent with the `( `?).  Here it was consistent and you remove the WS.

src/utils/flags/flagtypes.h
```diff
@@ -154,21 +169,25 @@ Value WriteFixedValueType ( T& target, const T &value ) {
 /**
  * Returns a \sa ZyppFlags::Value instance handling flags taking a string parameter representing a \sa zypp::filesystem::PathName
  */
-Value PathNameType( filesystem::Pathname &target, const boost::optional<std::string> &defValue, std::string hint );
+Value PathNameType( filesystem::Pathname &target, const boost::optional<std::string> &defValue = boost::optional<std::string>(), std::string hint = std::string() );
```
mlandres: JFYI: it's just `(zypp::)Pathname`. No one calls it   `(zypp::)filesystem::Pathname`.

PR #224, opened 2018-12-10
GlobalSettingSingleton: reset() should be the same as 'default constr…

mlandres:
…ucted', shouldn't it?  If so, it should IMO be implemented as ` = T()`.

The only Issue was `SolverSettingsData`. The default constructed TriBools were `false`, while `reset` explicitly assigned `indeterminate`. I suppose the later is what's actually wanted?


bzeller:
> …ucted', shouldn't it? If so, it should IMO be implemented as ` = T()`.
> 
> The only Issue was `SolverSettingsData`. The default constructed TriBools were `false`, while `reset` explicitly assigned `indeterminate`. I suppose the later is what's actually wanted?

Yes, whatever is in reset() is what's actually wanted. But it should never be used before reset() was called....

PR #227, opened 2018-12-13
Ma l3 locks

mlandres:
[ L3-Question: zypper locks display is broken](https://bugzilla.suse.com/show_bug.cgi?id=1112911).
```
type: package
match_type: glob
case_sensitive: on
solvable_name: a*
version: == 1

type: package
match_type: glob
case_sensitive: on
solvable_name: a* == 1

type: package
match_type: glob
case_sensitive: on
complex: AttrMatchData solvable:name a* C SolvableRange\ ==\ 1\ \"\"
```
What a mess. Real life `locks` files may contain the above 3 forms of versioned locks.

The 1st is the one zypper is actually able to handle in `list locks`.

The 2nd one is what zypper creates when creating a versioned lock. Unfortunately this query does not make any sense until libzypp reads it back from the file and translates it into a complex query (3rd). Side effect: You can't remove versioned lock 2nd and 3rd query are not `==`.

Semantically 1st and 3rd lock the same set, but zypper is unable to handle the 3rd form (neither in `list` nor in `remove locks`). 

The PR fixes zypper to create and correctly display the 1st form.

To fix already existing `locks` files containing the 2nd and 3rd form, libzypp will convert them to the 1st when parsing the file. After the next write, only the fist for should be present.

### DIFF
--- a/src/commands/locks/common.cc
+++ b/src/commands/locks/common.cc
@@ -1,0 +1,47 @@
+/*---------------------------------------------------------------------------*\
+                          ____  _ _ __ _ __  ___ _ _
+                         |_ / || | '_ \ '_ \/ -_) '_|
+                         /__|\_, | .__/ .__/\___|_|
+                             |__/|_|  |_|
+\*---------------------------------------------------------------------------*/
+/** \file commands/locks/common.cc
+ * Common code used by different commands.
+ */
+#include "commands/locks/common.h"
+#include "repos.h"
+
+///////////////////////////////////////////////////////////////////
+namespace locks
+{
+  zypp::PoolQuery arg2query( Zypper & zypper, const std::string & arg_r, const std::set<zypp::ResKind> & kinds_r, const std::vector<std::string> & repos_r )
+  {
+    PoolQuery q;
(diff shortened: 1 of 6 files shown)


src/commands/locks/common.cc
```diff
@@ -0,0 +1,100 @@
...
+    for_( it, repos_r.begin(), repos_r.end() )
+    {
+      RepoInfo info;
+      if ( match_repo( zypper, *it, &info ) )
+	q.addRepo( info.alias() );
+      else //TODO some error handling
+	WAR << "unknown repository" << *it << endl;
+    }
+
+    if ( kinds_r.empty() ) // derive it from the name
```
mlandres: This should be changed. If the arg contains an explicit kind like `pattern:foo`, then the query should also be built by the `if`. 
It makes no sense to add all kinds from the command line, but then overwrite this by adding an explicit search for `pattern:foo`.  A simple `kind=pattern` `name =foo` is enough.

bzeller: ah right, thats broken otherwise

PR #228, opened 2018-12-18
Refactor commands the final

bzeller:
This PR moves the last remaining commands over to the new style command objects. Also removes some now unused code.


mlandres:
```
-msgid "Print help."
+msgid "Print zypper help"

-msgid "Repository Options:"
+msgid "Repository Options"

-msgid "Target Options:"
+msgid "Target Options"
```
Please add the 3 old strings to `legacy_strings.cc`. I'll fix the translations later.

src/commands/configtest.cc
```diff
@@ -128,15 +113,23 @@ namespace
...
+ConfigTestCmd::ConfigTestCmd(const std::vector<std::string> &commandAliases_r):
+  ZypperBaseCommand (
+    commandAliases_r,
+    // translators: command synopsis; do not translate lowercase words
+    _("configtest"),
+    // translators: command summary: configtest
+    _("This command is for debuging purposes only."),
+    // translators: command description
+    _("This command is for debuging purposes only."),
+    DisableAll
```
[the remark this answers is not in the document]
bzeller: Done

src/commands/subcommand.cc
```diff
@@ -476,65 +476,23 @@ std::ostream & SubcommandOptions::showHelpOn( std::ostream & out ) const
...
-
-int subcommand( Zypper & zypper_r )
-{
-  zypper_r.cleanupForSubcommand();
-  return SubcommandImpl( zypper_r ).run();
-}
-
-int subcommand( Zypper & zypper_r, shared_ptr<SubcommandOptions> options_r )
-{
-  zypper_r.cleanupForSubcommand();
```
mlandres: Looks like the `zypper_r.cleanupForSubcommand();` call is lost (here and from L527). 
Should probably be added as 1st action to `SubCmd::runCmd`.

src/Zypper.cc
```diff
@@ -486,9 +230,15 @@ int Zypper::main( int argc, char ** argv )
...
-	  subcommand( *this );
+  case ZypperCommand::SUBCOMMAND_e: {
+    std::shared_ptr<SubCmd>  cmd = std::dynamic_pointer_cast<SubCmd>( command().commandObject() );
+    if ( cmd )
+      cmd->runCmd( *this );
+    else {
+      ZYPP_THROW( Exception ("Invalid command object") );
+    }
 	  break;
+  }
```
mlandres: Please indent the case block to match the other cases.

PR #230, opened 2019-01-04
Clearer message when candidate is from different vendor

krig:
When the newest package candidate is from a different vendor,
the current message to the user doesn't clarify which vendors
are involved. Adding the vendor names to the message makes
it easier for the user to decide what to do.

### DIFF
--- a/src/RequestFeedback.cc
+++ b/src/RequestFeedback.cc
@@ -120,8 +120,8 @@ std::string SolverRequester::Feedback::asUserString( const SolverRequester::Opti
       PoolItem highest = ui::asSelectable()(_objinst)->highestAvailableVersionObj();
       std::ostringstream cmdhint;
       cmdhint << "zypper install " << highest.asString();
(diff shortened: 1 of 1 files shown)


krig:
As a clearer motivation for this patch, this is what the message looks like when there is an upgrade available from another vendor, with the patch applied:

```
# zypper up syncthing
Loading repository data...
Reading installed packages...
There is an update candidate for 'syncthing' from vendor 'obs://build.opensuse.org/home:KGronlund', while the current vendor is 'openSUSE'. Use 'zypper install syncthing-1.0.0-lp150.153.1.x86_64' to install this candidate.
Resolving package dependencies...

Nothing to do.
# 
```

At least to me, seeing which vendor is providing the update is really valuable information in order to make a decision to allow the vendor change or not.

mlandres:
Thanks for the patch.

krig:
Great, thank you!

PR #232, opened 2019-01-08
Simplify the zypper application flow

bzeller:
This PR unifies and simplifies the zypper application flow. All commands are now handled the same way. Even subcommands and zypper shell.

### DIFF
--- a/src/utils/flags/zyppflags.cc
+++ b/src/utils/flags/zyppflags.cc
@@ -161,7 +161,7 @@ Value & Value::after( PostWriteHook &&postWriteHook )
 }
 
 
-int parseCLI( const int argc, char * const *argv, const std::vector<CommandGroup> &options, const int firstOpt )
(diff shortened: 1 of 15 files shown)


src/utils/flags/zyppflags.cc
```diff
@@ -299,7 +301,7 @@ int parseCLI( const int argc, char * const *argv, const std::vector<CommandGroup
   while ( true ) {
 
     int option_index = -1;      //index of the last found long option, same as in allOpts
-    int optc = getopt_long( argc, argv, shortopts.c_str(), longopts.data(), &option_index );
+    int optc = getopt_long( argc, argv , shortopts.c_str(), longopts.data(), &option_index );
```
mlandres: No change, just superfluous WS inserted

bzeller: Done

PR #237, opened 2019-01-22
Show reboot hint in zypper ps and summary (fixes bnc#1120263)

bzeller:


davidcassany:
@bzeller I have some concerns regarding this change. Returning a non zero value for correct installations and upgrades can potentially break any script using zypper. In fact this breaks KIWI builds and forces us to start parsing return codes in any zypper call. Also it returns reboot-needed even when bootstrapping.

Isn't there any other option for this feature than returning non zero values? See SUSE/kiwi#932

bzeller:
@davidcassany I see, will discuss that with @mlandres again once he is back. Generally that is the same behaviour as "zypper patch" has already , it will return that error code in case a patch is marked with "reboot needed".

mlandres:
> Generally that is the same behaviour as "zypper patch" has already 

OTOH we now have `zypper needs-rebooting` to check this anytime. We could leave `102/103` to the `patch` command and use `0` for *ordinary* commands.

davidcassany:
> OTOH we now have zypper needs-rebooting to check this anytime. We could leave 102/103 to the patch command and use 0 for ordinary commands.

That sounds reasonable to me, if I understand correctly this leaves zypper behavior as it was and it also includes the additional `needs-rebooting` command. That looks like good approach to me.

mlandres:
@davidcassany JFYI it's always been possible that zypper returns `102/103` if a `patch` carrying the `reboot/restart needed ` attribute got installed. New is only that some `packages` carry this attribute too.

davidcassany:
@mlandres @bzeller thanks for you feedback and cooperation. This makes sense to me, specially if it has been _always_ this way. So I understand there will be a change to not return 102/103 for _ordinary_ commands?

mlandres:
@davidcassany we will restore the previous behavior and return `102/103` only if `patches` are involved. Packages will cause visual feedback in the summary, but not influence the return code.

PR #252, opened 2019-02-08
Show only required info in the summary in quiet mode ( fixes bnc#993025)

bzeller:


src/Summary.cc
```diff
@@ -1548,9 +1548,11 @@ void Summary::dumpTo( std::ostream & out )
     unsetViewOption( SHOW_SUGGESTED );
   }
 
-  if ( _viewop & SHOW_LOCKS )
+  bool quiet = zypper.config().verbosity == Out::QUIET;
+
+  if ( !quiet && _viewop & SHOW_LOCKS )
     writeLocked( out );
-  if ( _viewop & SHOW_NOT_UPDATED )
+  if ( !quiet && _viewop & SHOW_NOT_UPDATED )
```
mlandres: I'd actually unset it in `_viewop`, to be sure in case any of the other `write*` functions would also refer to this option.
```
if ( zypper.config().verbosity == Out::QUIET )
{
  unsetViewOption( SHOW_LOCKS );
  unsetViewOption( SHOW_NOT_UPDATED );
}
```

bzeller: Done

PR #254, opened 2019-02-11
Move needs-restarting command to subpackage

Conan-Kudo:
This makes it so that dnf-utils and zypper can coexist on an
openSUSE system, which is important when dnf-utils is being used
on openSUSE with tools like Mock.

### DIFF
--- a/zypper.spec.cmake
+++ b/zypper.spec.cmake
@@ -111,6 +111,22 @@ Authors:
 --------
     Bernhard M. Wiedemann <bernhard+aptitude4zypp lsmod de>
 
+%package needs-restarting
+Summary:        needs-restarting compatibility with zypper
+Group:          System/Packages
+Requires:       zypper
+%if 0%{?suse_version}
+Supplements:    zypper
+%endif
+BuildArch:      noarch
+
+%description needs-restarting
+provides compatibility to YUM needs-restarting command using zypper
+
+Authors:
+--------
+    Michael Andres <ma@suse.de>
+
 %prep
 %setup -q
 
@@ -168,7 +184,6 @@ rm -rf "$RPM_BUILD_ROOT"
 %{_sysconfdir}/bash_completion.d/zypper.sh
 %{_bindir}/zypper
 %{_bindir}/yzpper
-%{_bindir}/needs-restarting
 %{_bindir}/installation_sources
 %{_sbindir}/zypp-refresh
 %dir %{_datadir}/zypper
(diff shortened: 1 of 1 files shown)


mlandres:
Need to check this with the SLE guys first, they requested to add the script. If it's moved, they may need to adapt their patterns/docs,etc.

Conan-Kudo:
@mlandres It should auto-install by default anyway, since it is supplementing the zypper package, but yeah...

sysrich:
I believe in Kubic/Transactional Server we need needs-restsarting, but we can’t rely on supplements ..

Conan-Kudo:
@sysrich Uhh, but needs-restarting is just a wrapper about `zypper needs-rebooting`, I'm pretty sure `transactional-update` uses the zypper subcommand.

Conan-Kudo:
@sysrich Based on what I see in https://github.com/openSUSE/transactional-update/commit/04e0c07bde1a1a76f43d132533579bf163fe6517, I don't think Kubic would break at all, since it doesn't use the `needs-restarting` shell wrapper script at all.

sysrich:
@conan-kudos and have you looked at https://github.com/SUSE/rebootmgr ?
Or kured?

Conan-Kudo:
@sysrich `rebootmgr` does not use it. Neither does `kured`.

mlandres:
AFAIKS `dnf-utils` does not include it, so where does it disturb?

Conan-Kudo:
@mlandres It does as of 4.0.0, which is currently in staging for TW and Leap 15.1.

Conan-Kudo:
For the moment, I've [purged the files to eliminate the conflict with zypper](https://build.opensuse.org/package/rdiff/system:packagemanager:dnf/dnf-plugins-core?linkrev=base&rev=10), because I can't test or use the package otherwise, but I really would rather have it. I've already prepared for when the `zypper-needs-restarting` package exists and added the necessary Conflicts to `dnf-utils`.

mlandres:
It's been an official feature request for 15.1, so I can't remove it until all stakeholders agree. The weak dependency bears the risk that things break if `dnf-utils` replaces `zypper-needs-restarting`. 
As I don't know who's using the script nor where it might have been documented, I asked for PMs opinion about this in the corresponding FATE request. Then we'll see....

Conan-Kudo:
@mlandres My understanding is that the zypper behavior would require user consent before swapping out `zypper-needs-restarting` for `dnf-utils` anyway. This is also true in DNF. So I would expect that in the SLE 15 SP1 case, it should have no meaningful impact, and for the openSUSE Leap 15.1 case, it should have negligible impact for most people.

That said, I can't see the FATE request, so I have no idea what the context was (it doesn't seem to exist in openSUSE FATE...).

Conan-Kudo:
@mlandres @bzeller ping?

mlandres:
@Conan-Kudo: Waiting for the release managers final 'OK'.

mlandres:
Looks like we can split it. Most probably for 1.14.26.

Conan-Kudo:
@mlandres So you can pull this PR in, then?

Issue #13, opened 2012-09-14
Add --ignore-unknown option for install/update


mlandres:
Requested 2012-09-14 by Juergen Gehne on [zypp-devel]:

> may I add something to the zypper wish list?
> 
> We very often install a group of packages via zypper automatically by
> 
> zypper --no-refresh --non-interactive --no-gpg-check  in pack1 pack2 ... packN
> 
> But if only 1 package cannot be found, zypper does not install any package.
> I would be very glad if you could implement an option like "--ignore-unknown" so that unknown
> packages will be ignored.

PR #261, opened 2019-02-22
fate#325599 config writing and the 'yes/no/always/never' prompt

mlandres:
Enhancements on the Augeas class to allow option writing. Also Enhancements to the prompt classes to support arbitrary translated words (not just single chars) as prompt reply (and not just 'yes/no').

### DIFF
--- a/src/utils/Augeas.cc
+++ b/src/utils/Augeas.cc
@@ -140,6 +140,9 @@ namespace
     const char * c_str() const
     { return _path.c_str(); }
 
+    explicit operator bool() const
+    { return !_path.empty(); }
+
     AugPath & operator=( const std::string & str_r )
     { str() = str_r; return *this; }
 
@@ -148,6 +151,16 @@ namespace
 
     AugPath operator+( const std::string & str_r ) const
     { return AugPath(*this) += str_r; }
+
+    /** Returns a unique AugPath to the first match or an empty AugPath if no matches.
+     * \see \ref AugMatches
+     */
+    AugPath firstMatch() const;
+
+    /** Returns a unique AugPath to the last match or an empty AugPath if no matches.
+     * \see \ref AugMatches
+     */
+    AugPath lastMatch() const;
 
   public:
     bool empty() const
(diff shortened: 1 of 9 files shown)


src/utils/prompt.cc
```diff
@@ -101,26 +120,76 @@ void PromptOptions::setOptionHelp( unsigned opt, const std::string & help_str )
...
 
-  unsigned i = 0;
-  for( StrVector::const_iterator it = _options.begin(); it != _options.end(); ++i, ++it )
+  // #NUM ? (direct index into option vector)
+  if ( reply_r[0] == '#' && reply_r[1] != '\0' )
   {
-    DBG << "index: " << i << " option: " << *it << endl;
-    if ( *it == str::toLower(reply) )
+    unsigned num = 0;	// -1 : if no match
+    for ( const char * cp = reply_r.c_str()+1; *cp; ++cp )
```
bzeller: for loop could stop once num gets bigger than what was expected, in case someone types a really large number

mlandres: The intention is actually to catch and eat `/^#[0-9]+$/` and let everything else fall through to the string matcher. So I have to parse it to the end and anyway. Discarding `#NUM` does not depend on whether `NUM` matches the number of options (otherwise one could name an option literaly `#13` and this would work until >=13 options are used).

src/utils/Augeas.cc
```diff
@@ -596,5 +651,55 @@ std::string Augeas::getOption( const std::string & option_r ) const
...
+
+  AugPath fP { _pimpl->augPath( "/files"/cfg/option_r ) };
+
+  unsigned matches = fP.matches();
+  if ( matches )
+  {
+    if ( matches > 1 )
+    {
+      AugPath last { fP.lastMatch() };
+      fP.rm();
```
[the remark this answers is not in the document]
mlandres: We could store the value back in the last location, but I wanted to disambiguate the file by removing the unwanted ones.

`rm` removes ALL locations matching the (ambiguous in this case) path and we store the value back at the (unique) last location.

mlandres:
> also as suggested replace PWD with getcwd()

Initially `yes`, but now uncertain because `PWD` remains unchanged if zypp does a `chdir()`. So it always contains what the user expected. `getcwd()` does not.

PR #266, opened 2019-03-04
Fix unknown package handling in zypper install (fixes bsc#1127608)

bzeller:
Fixes a regression in zypper install, which previously only stopped
on a unknown package argument when running in non-interactive mode.
This behaviour was broken when migrating the install command to
the new unified command object API, see commit f2e0f8638a32225ff084458a536eedc0f08ee549.

### DIFF
--- a/src/commands/installremove.cc
+++ b/src/commands/installremove.cc
@@ -24,7 +24,7 @@ void InstallRemoveBase::fillSrOpts(SolverRequester::Options &sropts_r ) const
   sropts_r.force_by_name = _selectByName;
 }
 
-int InstallRemoveBase::handleFeedback(Zypper &zypper, const SolverRequester &sr_r ) const
+void InstallRemoveBase::handleFeedback(Zypper &zypper, const SolverRequester &sr_r, bool failOnCapNotFound ) const
 {
   sr_r.printFeedback( zypper.out() );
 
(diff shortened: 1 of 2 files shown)


src/commands/installremove.cc
```diff
@@ -275,7 +275,12 @@ int InstallCmd::execute( Zypper &zypper, const std::vector<std::string> &positio
   sr.install( rpm_args );
 
   code = handleFeedback( zypper, sr );
-  if ( code != ZYPPER_EXIT_OK )
+  zypper.setExitCode( code );
+
+  //bsc#1127608 Only return with error if zypper is executed non interactively
+  //This was missed when migrating the install/remove commands,
+  //see commit f2e0f8638a32225ff084458a536eedc0f08ee549
+  if ( code != ZYPPER_EXIT_OK && zypper.config().non_interactive )
```
mlandres: The original workflow was similar to the flow still in in [update](https://github.com/openSUSE/zypper/blob/master/src/commands/update.cc#L140) or patch; throwing:

```
./Zypper.cc-
./Zypper.cc-    if ( !globalOpts().ignore_unknown
./Zypper.cc-      && ( sr.hasFeedback( SolverRequester::Feedback::NOT_FOUND_NAME )
./Zypper.cc-        || sr.hasFeedback( SolverRequester::Feedback::NOT_FOUND_CAP ) ) )
./Zypper.cc-    {
./Zypper.cc:      setExitCode( ZYPPER_EXIT_INF_CAP_NOT_FOUND );
./Zypper.cc-      if ( command () != ZypperCommand::REMOVE && globalOpts().non_interactive )    // bsc#980263: relax if removing packages
./Zypper.cc-        ZYPP_THROW( ExitRequestException("name or capability not found") );
./Zypper.cc-    }
```
I'd change `handleFeedback` to it exactly the same way (setExitCode and throw if non_interactive). Then add an arg `bool doThrow = true` replacing the above `command () != ZypperCommand::REMOVE`.

With this `install` just needs to call `handleFeedback( zypper, sr );` an `remove` does `handleFeedback( zypper, sr, false );`. (`handleFeedback` could even become `void`)

PR #270, opened 2019-03-15
Add option to show more conflict information

bzeller:
This patch adds a new option to the prompt generated in the
show_problem() function to show all available conflict information
to a specific solving problem.
Also cleans up the option handling a bit and adds help text.

Needs to land after: https://github.com/openSUSE/libzypp/pull/164

### DIFF
--- a/src/solve-commit.cc
+++ b/src/solve-commit.cc
@@ -151,9 +151,6 @@ static TriBool show_problem( Zypper & zypper, const ResolverProblem & prob, Prob
       fulltext << desc_ext_stm.str();
     fulltext << solution_stm.str();
 
-    if ( !zypper.config().non_interactive )
-      clear_keyboard_buffer();
-
     zypper.out().prompt( PROMPT_DEP_RESOLVE, prompt_text, popts, fulltext.str() );
     reply = get_prompt_reply( zypper, PROMPT_DEP_RESOLVE, popts );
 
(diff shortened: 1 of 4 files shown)


src/solve-commit.cc
```diff
@@ -87,47 +94,85 @@ static TriBool show_problem( Zypper & zypper, const ResolverProblem & prob, Prob
...
+    if ( !zypper.config().non_interactive )
+      clear_keyboard_buffer();
+
+    std::ostringstream fulltext;
+    fulltext << desc_stm.str();
+    if ( showExtInfo )
+      fulltext << desc_ext_stm.str();
+    fulltext << solution_stm.str();
+
+    zypper.out().prompt( PROMPT_DEP_RESOLVE, prompt_text, popts, fulltext.str() );
```
mlandres:
```
  if ( !zypper.config().non_interactive )
    clear_keyboard_buffer();
```
AFAICS almost all calls to `zypper.out().prompt` are _immediately_ preceded by the above. We should do it here too.

Or add a cleanup Commit? `grep -r -6 '\.prompt\>\|clear_keyboard_buffer' src/` shows that there are just a few cases where  `prompt` is called without `clear_keyboard_buffer`.  I suspect this is not intentional, but the call was simply forgotten. If so, we can move the preceding `clear_keyboard_buffer` call into `prompt` and remove them in the code.

bzeller: prompt() is a pure virtual, so we have 3 options:
A) should we move the clear_keyboard_buffer into each implementation
B) implement the clear_keyboard_buffer in the base and make each impl call it 
C) make a non virtual safePrompt() that calls clear_keyboard_buffer and then the virtual prompt()

Any favourites?

mlandres: Revisiting the code I'm not sure whether this is the right location at all. `Out::prompt` just prints the prompt while `clear_keyboard_buffer` should be done before reading the answer. So `src/utils/prompt.cc:get_prompt_reply:L361` (after opening the terminal) is probably a better place to clear ant typeahead.

bzeller: Moved clear_keyboard_buffer into get_prompt_reply and cleaned up all occurences right before a get_prompt_reply call

PR #292, opened 2019-09-26
Update zypper.8.txt

ralflang:
Clarify that locking also prevents upgrades. This has been misleading some to think it only prevents install/uninstall and there are multiple questions wrt locking in forums, stackoverflow etc.


mlandres:
Thanks for the patch.

PR #294, opened 2019-10-01
Ask only once if multiple packages share the same license text 

mlandres:
[bsc#1145554](https://bugzilla.opensuse.org/show_bug.cgi?id=1145554)

### DIFF
--- a/src/misc.cc
+++ b/src/misc.cc
@@ -99,13 +99,16 @@ namespace
 // - after negative answer, call solve_and_commit() again
 bool confirm_licenses( Zypper & zypper )
 {
-  bool confirmed = true;
+  std::unordered_map<std::string,std::set<PoolItem>> licenseCollector; // bsc#1145554: collect multiple packages using the same license text
+
   bool auto_agree_all = LicenseAgreementPolicy::instance()._autoAgreeWithLicenses;
   bool auto_agree_product = auto_agree_all || LicenseAgreementPolicy::instance()._autoAgreeWithProductLicenses;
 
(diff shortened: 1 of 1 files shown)


src/misc.cc
```diff
@@ -99,13 +99,16 @@ namespace
 // - after negative answer, call solve_and_commit() again
 bool confirm_licenses( Zypper & zypper )
 {
-  bool confirmed = true;
+  std::map<std::string,std::set<PoolItem>> licenseCollector; // bsc#1145554: collect multiple packages using the same license text
```
bzeller: we could use a unordered_map , or do we care about preserving the order?

src/misc.cc
```diff
@@ -140,62 +143,81 @@ bool confirm_licenses( Zypper & zypper )
...
 
-      bool to_accept = pi.needToAcceptLicense();	// true except for e.g. openSUSE which wants the text to be shown, but no need to agree.
-      if (to_accept)
+  // now display them...
+  for ( auto el : licenseCollector )
+  {
+    std::ostringstream s;
+    bool toAccept = false;	// e.g. openSUSE which wants the text to be shown, but no need to agree.
+
+    for ( const PoolItem & pi : el.second )
```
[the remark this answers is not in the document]
mlandres: Force of habit and not worse than `auto`.

src/misc.cc
mlandres: No, we could use an unordered_map as well (there is no order to preserve).

PR #296, opened 2019-10-09
Add --repo switch to download

bzeller:
Introduces the --repo switch to the zypper download command so
downloading can be limited to a specific repository (jira#SLE-9171)

### DIFF
--- a/src/commands/utils/download.cc
+++ b/src/commands/utils/download.cc
@@ -166,7 +166,11 @@ ZyppFlags::CommandGroup DownloadCmd::cmdOptions() const
          ZyppFlags::BoolType( &that->_allMatches, ZyppFlags::StoreTrue, _allMatches ),
          // translators: --all-matches
          _("Download all versions matching the commandline arguments. Otherwise only the best version of each matching package is downloaded.")
-      }
+      },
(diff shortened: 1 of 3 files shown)


src/commands/utils/download.cc
```diff
@@ -166,7 +166,11 @@ ZyppFlags::CommandGroup DownloadCmd::cmdOptions() const
          ZyppFlags::BoolType( &that->_allMatches, ZyppFlags::StoreTrue, _allMatches ),
          // translators: --all-matches
          _("Download all versions matching the commandline arguments. Otherwise only the best version of each matching package is downloaded.")
-      }
+      },
+      { "from", '\0', ZyppFlags::Repeatable | ZyppFlags::RequiredArgument, ZyppFlags::StringVectorType( &InitRepoSettings::instanceNoConst()._repoFilter, ARG_REPOSITORY),
+        // translators: --from <ALIAS|#|URI>
+        _("Work only with the specified repository.")
```
mlandres: I'd prefer to keep a different wording for `--repo` and `--from`, even if the behavior is similar/same.

For `--repo` AFAIK  `Work only with the specified repository.` is used (which translates to  'only these repos are loaded into the pool').

For `--from` something like `Select packages from the specified repository.` (here using --repo in the implementation enforces this, but it's not an option for all commands).

PR #301, opened 2019-10-29
Always execute commit when adding/removing locales (fixes bsc#1155205)

bzeller:


src/solve-commit.h
```diff
@@ -26,16 +26,26 @@
 bool resolve(Zypper & zypper);
 
 
+/**
+ * Defines the \sa solve_and_commit commit policy.
+ * Usually \sa solve_and_commit only executes a commit if packages need to
```
mlandres: Inside a sentence you probably want `\ref` (reference link) rather than `\sa` (__See also__ paragraph). There are some more in zypper and I also found some in libzypp. We should do fixup commits for them.

Look how it's rendered: [` *  Returns the same as \sa systemRoot() if not explicitely set.`](https://doc.opensuse.org/projects/libzypp/HEAD/classzypp_1_1ZConfig.html#a2666b9ad701612bd53fb7a0c16ebd1cc)
And this is how `\ref` looks like: [`/** Set a new \ref userData string.`](https://doc.opensuse.org/projects/libzypp/HEAD/classzypp_1_1ZConfig.html#a5804ba6191a70e3dbed9d22105acb5fc)

bzeller: Done

PR #304, opened 2019-11-07
Port zypper-log to python3

asdil12:
Bug: https://bugzilla.suse.com/show_bug.cgi?id=1156158

### DIFF
--- a/tools/zypper-log
+++ b/tools/zypper-log
@@ -1,9 +1,28 @@
-#!/usr/bin/env python
+#!/bin/sh
 
 # This tool helps you accessing the zypper logfile
 # see --help for more details
 #
 # Author: Dominik Heidler <dheidler@suse.de>
+
+# Shell commands follow
+# Next line is bilingual: it starts a comment in Python, and is a no-op in shell
+""":"
+
+if which python3 > /dev/null 2>&1 ; then
+	PYTHON=python3
+elif which python > /dev/null 2>&1 ; then
+	PYTHON=python
+elif which python2 > /dev/null 2>&1 ; then
+	PYTHON=python2
+fi
+
+exec $PYTHON $0 "$@"
+
+":"""
+# Previous line is bilingual: it ends a comment in Python, and is a no-op in shell
+# Shell commands end here
+# Python script follows
 
 from __future__ import print_function
 import os, string, re, bz2, zlib, sys, time, argparse, errno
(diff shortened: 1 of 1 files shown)


mlandres:
Can it be done to work with 2 and 3? The script is shipped down to SLE15.

asdil12:
Wouldn't it be easier to just not backport it?
Or is that done automatically?

asdil12:
Bacause otherwise I would still need to rely on `xzcat` for uncompressing the logs as the `lzma` module is only added in python3.

asdil12:
So this code would run on both python2 and 3: https://github.com/asdil12/zypper/commit/1338d8f502fcdf22ca76114ccfd139f1ff6fb5fc

BUT: On SLE15SP2 there is no python binary - only python3 for python3 as python2 is still installable.

So either we add a dependency to python in the zypper rpm, so that it pulls python2 providing the python binary, or we change the shebang to python3.

mlandres:
> Wouldn't it be easier to just not backport it?

SLE15* is all the same sourcecode. 

> or we change the shebang to python3.

We can use `sh` to test whether /usr/bin/python or python3 is there and feed the pycode as a here-document:
```
#! /bin/sh
PYTHON=....
$PYTHON <<__EOF__
...
__EOF__
```

asdil12:
Done.

mlandres:
Blame python ;)
Danke!

Conan-Kudo:
Why didn't you just do a sed in the spec file to change the shebang? That's what I do in Fedora downstream: https://src.fedoraproject.org/rpms/zypper/blob/42c0dd08551fbdd30aa0429c503c48d8eb07a356/f/zypper.spec#_106-107

Conan-Kudo:
I've proposed a far less insane solution: https://github.com/openSUSE/zypper/pull/306

mlandres:
The script needs to run with py2 and 3.

Conan-Kudo:
@mlandres Not on SLE 15. Python 3 has been available forever, so there it can always run on Python 3.

DimStar77:
This is indeed quite 'dirty' - you could just enforce python3 even down to SLE12, wherepython3 has been available.

The current shebang again avoids rpm from adding proper dependencies on an actual python interpreter - it's still relying on python 'just being there'

The approach of 'just going py3', which is supportable down to SLE12GA, would avoid that. Yu could add /usr/bin/oython3 as shebang and rpm would add this binary as a dependency - allowing to have an actually working package in all situations

Conan-Kudo:
@DimStar77 I've proposed as such in #306.

fcrozat:
> @DimStar77 I've proposed as such in #306.

I would suggest to alter it to use py2 on SLE12 codebase (py3 is not by default there) and py3 everywhere else.

Doing this runtime check is way too complex for no gain at all.

mlandres:
Too much ado. Will be replaced by a bash script. getopt and sed are sufficient.

Conan-Kudo:
> > @DimStar77 I've proposed as such in #306.
> 
> I would suggest to alter it to use py2 on SLE12 codebase (py3 is not by default there) and py3 everywhere else.
> 
> Doing this runtime check is way too complex for no gain at all.

Done. I've pushed it so that it uses Python 2 for older than SLE 15.

Conan-Kudo:
> > @DimStar77 I've proposed as such in #306.
> 
> I would suggest to alter it to use py2 on SLE12 codebase (py3 is not by default there) and py3 everywhere else.
> 
> Doing this runtime check is way too complex for no gain at all.

@fcrozat Done. I've pushed it so that it uses Python 2 for older than SLE 15.

PR #306, opened 2019-11-13
Set the shebang properly for zypper-log

Conan-Kudo:
Doing insane things like having a shell stub to locate a Python to
execute causes problems, particularly in scenarios where no matching
Python interpreter was installed.

Instead, let's do the reasonable thing and actually just set the
correct shebang in the spec file so that it works properly when packaged.

This solution was adapted from Fedora's zypper package: https://src.fedoraproject.org/rpms/zypper/c/1eeac978cbea6494705e3f5785c12de458a1aa2b

### DIFF
--- a/tools/zypper-log
+++ b/tools/zypper-log
@@ -1,28 +1,9 @@
-#!/bin/sh
+#!/usr/bin/env python
 
 # This tool helps you accessing the zypper logfile
 # see --help for more details
 #
 # Author: Dominik Heidler <dheidler@suse.de>
-
-# Shell commands follow
-# Next line is bilingual: it starts a comment in Python, and is a no-op in shell
-""":"
-
-if which python3 > /dev/null 2>&1 ; then
-	PYTHON=python3
-elif which python > /dev/null 2>&1 ; then
-	PYTHON=python
-elif which python2 > /dev/null 2>&1 ; then
-	PYTHON=python2
-fi
-
-exec $PYTHON $0 "$@"
-
-":"""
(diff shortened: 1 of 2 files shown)


mlandres:
If it should work with py3 only, we'd changed the shebang. The script needs to run with py2 and 3.

Conan-Kudo:
@mlandres I could have easily set it to work with Python 2 as well. But there's no reason for it to be called by Python 2 on SLE 15.

Conan-Kudo:
@mlandres The script will work with either (which is why the unversioned shebang), but the packaging should _make a choice_. That's why it's set up this way.

mlandres:
Why should I make a choice. I don't mind if a SLE15 runs py2 or 3.

Conan-Kudo:
@mlandres Because Python 2 is upstream EOL in a month and a half, and the lifetime for SLE (and core packages in SLE) is a heck of a lot longer than that. And while I can't read the bug report in the SUSE/Novell Bugzilla, I can guess that the complaint was that it wasn't running on SLE 15 because by default there is no `/usr/bin/python`. It's perfectly fine for the upstream codebase to support running on Python 2, but for these reasons and more, it should be running using Python 3 downstream in SLE 15.

And when SLE 16 rolls around, `/usr/bin/python` will point to Python 3 anyway (hopefully!), like it does in Fedora 31 now.

fcrozat:
there is no point in supporting py2 on SLE15.

Conan-Kudo:
@fcrozat @DimStar77 For some reason, this PR is not updating anymore even though I've updated the branch...

PR #308, opened 2019-11-14
Purge kernels

bzeller:
Adds a new zypper command to cleanup all obsolete kernels as configured by the user.

### DIFF
--- a/doc/zypper.8.txt
+++ b/doc/zypper.8.txt
@@ -436,7 +436,7 @@ This command also accepts the *Download-and-install mode options* described in t
 		Select packages by capabilities.
 
 	*-D*, *--dry-run*::
-		Test the removal of packages, do not actually remove anything. If used together with *--download-only* a meaningful file conflict check can be performed (see section *Package File Conflicts*).
+		Test the removal of packages, do not actually remove anything.
 
 	*--details*::
 		Show the detailed installation summary.
@@ -453,6 +453,19 @@ include::{incdir}/option_Solver_Flags_Common.txt[]
 
 	*-U*, *--no-clean-deps*::
 		No automatic removal of unneeded dependencies.
+--
+
+*purge-kernels* ['options']::
+	Autoremoves installed kernels.
+	+
+	Automatically cleans up installed kernels according to the rules defined in [zypp.conf:multiversion.kernels] which can be given as <version>, latest(-N), running, oldest(+N).
++
+--
+	*--details*::
+		Show the detailed installation summary.
+
+	*-D*, *--dry-run*::
+		Test the removal of packages, do not actually remove anything.
 --
 
 Update Management Commands
--- a/src/commands/utils/purge-kernels.cc
+++ b/src/commands/utils/purge-kernels.cc
@@ -1,0 +1,75 @@
+/*---------------------------------------------------------------------------*\
+                          ____  _ _ __ _ __  ___ _ _
+                         |_ / || | '_ \ '_ \/ -_) '_|
+                         /__|\_, | .__/ .__/\___|_|
+                             |__/|_|  |_|
+\*---------------------------------------------------------------------------*/
+
+#include "purge-kernels.h"
+#include "commands/conditions.h"
+#include "commands/commonflags.h"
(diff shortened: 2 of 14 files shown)


src/commands/utils/purge-kernels.cc
```diff
@@ -0,0 +1,67 @@
...
+PurgeKernelsCmd::PurgeKernelsCmd(std::vector<std::string> &&commandAliases_r) :
+  ZypperBaseCommand(
+    std::move( commandAliases_r ),
+    // translators: command synopsis; do not translate lowercase words
+    _("purge-kernels [OPTIONS]"),
+    // translators: command summary: targetos, tos
+    _("Remove old kernels."),
+    // translators: command description
+    _("Autoremoves installed kernels accoring to list of kernels to keep from /etc/zypp/zypp.conf:multiversion.kernels which can be given as <version>, latest(-N), running, oldest(+N)."),
+    ResetRepoManager | InitTarget | LoadTargetResolvables | Resolve
```
mlandres: Shall we introduce shortcuts? `PackageRemoveSetup = ResetRepoManager | InitTarget | LoadTargetResolvables | Resolve`, also `PackageInstallSetup`

doc/zypper.8.txt
```diff
@@ -455,6 +455,19 @@ include::{incdir}/option_Solver_Flags_Common.txt[]
 		No automatic removal of unneeded dependencies.
 --
 
+*purge-kernels* ['options']::
+	Autoremoves installed kernels.
+	+
+	Automatically cleans up installed kernels according to the rules defined in /etc/zypp/zypp.conf:multiversion.kernels which can be given as <version>, latest(-N), running, oldest(+N).
```
mlandres: The AFAICS most common way of referring to a zypp.conf entry in the man page is `[zypp.conf:multiversion.kernels]`.

doc/zypper.8.txt
```diff
@@ -455,6 +455,19 @@ include::{incdir}/option_Solver_Flags_Common.txt[]
...
++
+--
+	*--details*::
+		Show the detailed installation summary.
+
+	*-D*, *--dry-run*::
+		Test the removal of packages, do not actually remove anything. If used together with *--download-only* a meaningful file conflict check can be performed (see section *Package File Conflicts*).
+--
```
mlandres: Here and in the `remove` command we should IMO remove the `If used together with....` sentence. Remove has no download-only. It's somewhat confusing here.

src/commands/utils/purge-kernels.cc
```diff
@@ -0,0 +1,67 @@
...
+}
+
+void PurgeKernelsCmd::doReset()
+{}
+
+int PurgeKernelsCmd::execute( Zypper &zypper , const std::vector<std::string> & )
+{
+  PurgeKernels purger;
+  purger.markObsoleteKernels();
+
```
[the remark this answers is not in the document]
bzeller: Done

zypper.spec.cmake
```diff
@@ -26,7 +26,7 @@ BuildRequires:  boost-devel >= 1.33.1
 BuildRequires:  cmake >= 2.4.6
 BuildRequires:  gcc-c++ >= 4.7
 BuildRequires:  gettext-devel >= 0.15
-BuildRequires:  libzypp-devel >= 17.15.1
+BuildRequires:  libzypp-devel >= 17.16.1
 BuildRequires:  readline-devel >= 5.1
```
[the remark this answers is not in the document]
bzeller: Done

doc/zypper.8.txt
bzeller: Done

doc/zypper.8.txt
bzeller: Done

src/commands/utils/purge-kernels.cc
bzeller: We could add that, but its not going to be used a lot , if you check out the code for the remove command the init is much more customized, e.g splitted in 2 parts and also uses "LoadResolvables" instead of "LoadTargetResolvables"

PR #311, opened 2019-11-18
zypper-log: Replace python by a bash script

mlandres:
(fixes#304, fixes#306, [bsc#1156158](https://bugzilla.suse.com/show_bug.cgi?id=1156158))

### DIFF
--- a/tools/zypper-log
+++ b/tools/zypper-log
@@ -1,171 +1,252 @@
-#!/bin/sh
+#!/bin/bash
+COMMAND="$(basename "$0")"
+ZYPPERLOG="/var/log/zypper.log"
 
-# This tool helps you accessing the zypper logfile
-# see --help for more details
-#
-# Author: Dominik Heidler <dheidler@suse.de>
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
This is an incredibly horrifying script. I can't see this being reasonably maintainable...

You wrote this just because you aren't willing to deal with a Python shebang?

(cc: @DimStar77, @fcrozat)

mlandres:
> This is an incredibly horrifying script. I can't see this being reasonably maintainable...

That's exactly what I thought when looking at the python script. 

> You wrote this just because you aren't willing to deal with a Python shebang?

No. I wrote it to get rid of the python dependency which is actually not needed for a script that basically just greps a few lines out of a log file. 

At the time Dominik filed his PR, the plan to replace it by  a bash script was already made. Dominiks solution would have allowed me to ship it to existing SLE15 systems even if they do not have py3. Resolving bsc#1156158 immediately. Making the script less urgent.

PR #316, opened 2019-11-28
Use default package cache location for temporary repos (fixes bsc#113…

bzeller:
…0873)

### DIFF
--- a/src/Zypper.cc
+++ b/src/Zypper.cc
@@ -555,7 +555,10 @@ void Zypper::doCommand( int cmdArgc, char **cmdArgv, int firstFlag )
             repo.setName( url.asString() );
 
             repo.setMetadataPath( runtimeData().tmpdir / repo.alias() / "%AUTO%" );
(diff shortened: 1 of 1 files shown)


src/Zypper.cc
```diff
@@ -555,7 +555,7 @@ void Zypper::doCommand( int cmdArgc, char **cmdArgv, int firstFlag )
             repo.setName( url.asString() );
 
             repo.setMetadataPath( runtimeData().tmpdir / repo.alias() / "%AUTO%" );
-            repo.setPackagesPath( Pathname::assertprefix( _config.root_dir, ZYPPER_RPM_CACHE_DIR ) );
+            repo.setPackagesPath( Pathname::assertprefix( _config.root_dir, _config.rm_options.repoPackagesCachePath ) / "@System" );
```
mlandres: Maybe a comment that @System was chosen because libzypp won't autoclean it (i.e. a different dirname won't work without touching libzypp)

PR #318, opened 2019-12-10
Make Out::zyppExceptionReport show exception history

bzeller:


mlandres:
Are you going to add a `out()error( e.asUserString() ) ` -> `zypper.out().error( e )` commit? IMO a unified output handling can't be bad. We'll probably get some reports about the quality of the history entries, but this is somethign we can enhance then.

mlandres:
Fixes #317

PR #319, opened 2019-12-11
according instead of accoring

skriesch:
I translated the sentence "Autoremoves installed kernels accoring to list of kernels to keep from /etc/zypp/zypp.conf:multiversion.kernels which can be given as <version>, latest(-N), running, oldest(+N)."

The word "accoring" does not exist in English. That should be "according". That has been changed with sed in all files.


mlandres:
Danke!

PR #322, opened 2020-01-23
Unify computation of the status indicator in several outputs

bzeller:


src/utils/misc.cc
```diff
@@ -698,3 +699,75 @@ void packagekit_suggest_quit()
...
+  bool installed = false;
+
+
+  Zypper &zypper = Zypper::instance();
+  const auto &repos = zypper.runtimeData().repos;
+
+  if ( traits::isPseudoInstalled( s.kind() ) )
+    installed = s.theObj().isSatisfied();
+  // check for installed counterpart in one of specified repos (bnc #467106)
+  else if ( InitRepoSettings::instance()._repoFilter.size() )
```
bzeller: @mlandres Do we need to move that code into Selectable::identicalInstalled as well?

mlandres:
fixed by #330

PR #323, opened 2020-02-03
zypper-aptitude: don't supplement zypper

lnussel:
supplementing zypper means zypper-aptitude gets installed by default
and pulls in perl. Neither is desired on small systems.

### DIFF
--- a/zypper.spec.cmake
+++ b/zypper.spec.cmake
@@ -95,9 +95,6 @@ Summary:        aptitude compatibility with zypper
 Group:          System/Packages
 Requires:       perl
 Requires:       zypper
-%if 0%{?suse_version}
-Supplements:    zypper
-%endif
 BuildArch:      noarch
 
 %description aptitude


bmwiedemann:
There is also some spec syntax to have it only pulled in where there is perl and zypper - would that be better?

zypper.spec.cmake
```diff
@@ -95,9 +95,6 @@ Summary:        aptitude compatibility with zypper
 Group:          System/Packages
 Requires:       perl
 Requires:       zypper
-%if 0%{?suse_version}
-Supplements:    zypper
```
Conan-Kudo: Change this to `Supplements: (zypper and perl)` if you only want it pulled in when Perl is already installed.

Alternatively, you could do `Supplements: (zypper unless openSUSE-MicroOS-release)` to prevent it from supplementing on MicroOS systems.

lnussel: hardcoding that logic into packages makes it hard to create derivatives. We have patterns as central place to control what's installed by default instead.

PR #324, opened 2020-02-21
Do not allow the abbreviation of cli arguments (bsc#1164543)

bzeller:


DimStar77:
Will this mean the often used and documented "zypper in --no-r" won't work anymore?

DimStar77:
at least transactional-updates (and thus possibly also other consumers) rely on --xml (short for --xmlout)

bzeller:
> Will this mean the often used and documented "zypper in --no-r" won't work anymore?

Yes, all abbreviations will stop working, we decided to change that to not run into name clashes when introducing new switches. For example --reposd-dir did take arguments for --repo , which is problematic if we ever need a global --repo switch.

bmwiedemann:
Will `zypper in --no-recommends` still work, or only `zypper install --no-recommends` ?

bzeller:
> Will `zypper in --no-recommends` still work, or only `zypper install --no-recommends` ?

zypper in will still work. Thats only about the switches

PR #325, opened 2020-03-02
Support --disable-repos as abbreviation

bzeller:


src/Config.cc
```diff
@@ -575,15 +583,15 @@ std::vector<ZyppFlags::CommandGroup> Config::cliOptions()
...
-                MIL << "Repositories disabled, using target only." << endl;
-                Zypper::instance().out().info(
-                  _("Repositories disabled, using the database of installed packages only."),
-                  Out::HIGH);
-              })),
+              .after([ notifyDisableRepos ](){ notifyDisableRepos(); })),
               // translators: --disable-repositories
               _("Do not read meta-data from repositories.")
         },
+        // synonym for --disable-repositories for SUMA
```
mlandres: Please add a thet this can be reverted after SUMA updated their SALT stack.

bzeller: Done

PR #327, opened 2020-03-04
Correctly detect ambigous switch abbrev / Still allow legacy longopt abbrev on 15.0/15.1

mlandres:
[bsc#1164543](https://bugzilla.suse.com/1164543)
[bsc#1165573](https://bugzilla.suse.com/1165573)
Includes and fixes #326 

On 15.0/15.1 just issue a warning if abbrev. switches are used.

### DIFF
--- a/src/utils/flags/zyppflags.cc
+++ b/src/utils/flags/zyppflags.cc
@@ -63,7 +63,15 @@ namespace {
 
     //we do not use the flag and val types, instead we use optind to figure out what happend
     using OptType = struct option;
-    opts.push_back( OptType{opt.name.c_str(), has_arg, 0 ,0} );
+
+    opts.push_back( OptType{
+      opt.name.c_str(),
+      has_arg,
+      0 ,
+      // set the "val" member to some unique number to allow a correct detection of a ambigous switch name (bsc#1165573)
+      // make sure the number is negative so we can differ between short and long options in case of error
+      - 100 - static_cast<int>(opts.size())
+    });
   }
 }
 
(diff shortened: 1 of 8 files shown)


src/utils/flags/zyppflags.cc
```diff
@@ -365,10 +383,16 @@ int parseCLI( const int argc, char * const *argv, const std::vector<CommandGroup
 
             zypp::str::smatch what;
             if ( zypp::str::regex_match( scanned, what, rxexpr ) && what.size() >= 2 ) {
-              if( opt.name != what[1] )
-                ZYPP_THROW( UnknownFlagException( scanned ) );
+              if( opt.name != what[1] ) {
+                if ( onlyWarnOnAbbrevSwitches() ) {
+		  WAR << "Abbreviated commandline option will not be supported in future versions: --" << what[1] << ": use --" << opt.name << endl;
```
bzeller: Shouldn't we use the plural of option here:  "Abbreviated commandline options will not ..." , currently it could be understood that only the currently used one won't be supported anymore

src/utils/messages.cc
```diff
@@ -118,6 +118,12 @@ std::string legacyCLIStr( const std::string & old_r, const std::string & new_r,
        _("Legacy commandline option %1% detected. This option is ignored."))
        % NEGATIVEString(dashdash(old_r));
     break;
+  case LegacyCLIMsgType::Abbreviated:
+    return str::Format(
+       _("Abbreviated commandline option %1% will not be supported in future versions. Please use %2% instead.") )
```
bzeller: Same as above, restructuring the sentence to use the plural of "option" would be less ambigous

PR #349, opened 2020-08-16
Remove extern C block wrapping augeas.h

DaanDeMeyer:
Breaks the build on Arch Linux with "error: template with C linkage".


mlandres:
Thanks.

PR #352, opened 2020-08-26
Directly list subcommands in 'zypper help'

mlandres:
[bsc#1165424](https://bugzilla.opensuse.org/show_bug.cgi?id=1165424)

SubCommand summary is based on the man page every subcommand should provide (`man -f`)


src/commands/subcommand.cc
```diff
@@ -40,74 +41,65 @@ namespace env
...
-#else
-      // horizontal
-      for ( unsigned idx = 0; idx < names_r.size(); ++idx )
-      {
-	if ( idx && idx % maxcols == 0 )
-	  str << endl;
-	str << ( fmter % names_r[idx] );
+      else {
+	// translators: %1% is the name of the command which has no man page available.
+	static str::Format fmt( "<"+ LOWLIGHTString(_("No manual entry for %1%")).str() + ">" );
```
bzeller, line 79: Why is this static? You are changing the variable in the next line, leaving it changed for the next time that code is called. Doesn'T that change the behaviour?

src/commands/subcommand.cc
```diff
@@ -144,37 +136,78 @@ namespace	// subcommand detetction
...
     for ( const auto & dir : dirs )
     {
-      collectSubommandsIn( dir, std::inserter(pathCommands_r,pathCommands_r.end()) );
+      collectSubommandsIn( dir );
     }
   }
 
-  inline void collectAllSubommands( std::set<std::string> & allCommands_r )
-  { collectAllSubommands( allCommands_r, allCommands_r ); }
+  /* The command details for the long help. */
```
bzeller, line 188: Does this comment belong here?

src/commands/subcommand.cc
```diff
@@ -144,37 +136,78 @@ namespace	// subcommand detetction
...
+			      if ( ! cmd._cmd.empty() )
+				fnc_r( std::move(cmd) );
 			      return true;
 			    } );
-    return cnt;
   }
 
-  inline void collectAllSubommands( std::set<std::string> & execdirCommands_r, std::set<std::string> & pathCommands_r )
+  /* Just the command names for the short help. */
+  inline void collectAllSubommands( std::set<std::string> & allCommands_r )
```
bzeller, line 168: collectAllSubommands -> collectAllSubCommands

bzeller: Also its a bit confusing to overload that function later. This is not really collecting all the subcommands but only the names of them. Should the function be renamed to reflect that?

src/commands/subcommand.cc
```diff
@@ -144,37 +136,78 @@ namespace	// subcommand detetction
...
   {
-    collectSubommandsIn( SubcommandOptions::_execdir, std::inserter(execdirCommands_r,execdirCommands_r.end()) );
+    auto collectSubommandsIn = [&allCommands_r]( const Pathname & dir_r ) {
```
bzeller, line 170: collectSubommandsIn -> collectSubCommandsIn

bzeller (changes requested):

src/commands/subcommand.cc
mlandres: No. The format string itself does not change anymore. It can be reused in subsequent substitutions. Just the state of the internal stream object that builds the substituted string changes when placeholder arguments are fed.

src/commands/subcommand.cc
mlandres: done

src/commands/subcommand.cc
mlandres: collectAllSubCommandNames

src/commands/subcommand.cc
mlandres: Yes. `/* Just the command names for the short help. */` vs `/* The command details for the long help. */`.

src/commands/subcommand.cc
mlandres: In fact most of the code spells it `Subcommand`, so I changed all the `SubCommand`

src/commands/subcommand.cc
```diff
@@ -40,74 +41,65 @@ namespace env
...
+  inline CommandSummaries getCommandsummaries( const DetectedCommands & commands_r )
+  {
+    CommandSummaries ret;
+    for ( auto & cmd : commands_r ) {
+      std::string sum { ExternalProgram( "man -f "+cmd._name ).receiveLine() };
+      if ( ! sum.empty() ) {
+	// # man -f zypper
+	// zypper (8)           - Command-line interface to ZYpp system management library (libzypp)
+	static const std::string sep { " - " };
+	std::string::size_type pos = sum.find( sep );
```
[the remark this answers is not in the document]
mlandres: Could, but I don't think this will make any difference. Micro optimization is done by the compiler. No matter if I write const or not, he'll has to check that the local variable is actually never modified before optimizing.

bzeller (approved):
OK

PR #362, opened 2020-10-26
Avoid translated text in xml attributes ( fixes #361 )

### LINKED ISSUE #361: Arguments translated in XML output
We had a probelm with an interaction between Zypper and Ansible. 

As you can read in the issue ansible-collections/community.general#1168, the Ansible module `zypper` use the XML output of `zypper` command to guess if a package is present or not. It looks for the attribute `name`. Unfortunately, if the LANG is something different than English, the attributes in the XML output are translated. 

Here is an sample of XML output with '_en_US_' locales and with '_fr_FR_':
- **en_US**: `<solvable status="installed" name="libwbclient0" kind="package" edition="4.7.6+git.54.6e3276c9872-2.47" arch="x86_64" repository="PTS-SLES-15_basesystem-x86_64"/>`
- **fr_FR**: `<solvable status="installed" nom="libwbclient0" kind="paquet" edition="4.7.6+git.54.6e3276c9872-2.47" arch="x86_64" dépôt="PTS-SLES-15_basesystem-x86_64"/>`

Don't you think it's problematic to have XML attribute changing with LANG? I d'ont feel confortable with that because at least several applications may not expect that.

bzeller:


mlandres (approved):
Ok

PR #366, opened 2020-12-23
Extend apt packagemap

bmwiedemann:
Extend apt-packagemap
based on the apt line in
https://docs.yoctoproject.org/brief-yoctoprojectqs/index.html

debianutils does not have a proper equivalent for
    run-parts
    tempfile
    installkernel
    ischroot
    savelog
    add-shell
    remove-shell


mlandres:
Thanks.

PR #372, opened 2021-01-19
Suggest scout version of command-not-found on zypper based systems

hellcp:
Associated with https://build.opensuse.org/request/show/864365 as we want to keep zypper-based systems using scout-based cnf, while dnf will use PackageKit-based one by default


mlandres:
IMO something for the distro (e.g. the -release package) to decide. Zypper actually has no opinion about the tool.

hellcp:
Yeah, I guess I will make `PackageKit-command-not-found` non-installable with zypp backend instead

PR #378, opened 2021-02-25
download cmd issues

mlandres:


src/commands/basecommand.h
```diff
@@ -237,6 +237,12 @@ class ZypperBaseCommand
    */
   virtual void doReset () = 0;
 
+  /**
+   * Reimplement for positionalArgs checks which can be done before systemSetup.
+   * Default implemntation does nothing.
```
bzeller, line 242:
```suggestion
   * Default implementation does nothing.
```

bzeller (approved):
After the typo is fixed it can be merged

src/commands/basecommand.h
mlandres: done

PR #385, opened 2021-04-30
patch: Let patch reboot-needed flag overrule included packages 

mlandres:
[bsc#1183268](https://bugzilla.suse.com/show_bug.cgi?id=1183268)

### DIFF
--- a/src/Summary.cc
+++ b/src/Summary.cc
@@ -145,6 +145,22 @@ void Summary::readPool( const ResPool & pool )
     }
   }
 
+  if ( hasViewOption( PATCH_REBOOT_RULES ) ) {
+    // bsc#1183268: Patch reboot-needed flag overrules included packages.
+    // Work around the unfortunate case, where a maintenance patch's metadata
+    // does not request a reboot, while some included package does.
+    // The 'patch --skip-interactive' command operates based on the maintenance
+    // metadata and skips patches that require a reboot. In this case we suppress
+    // the reboot needed hint of included packages in order not to confuse the user.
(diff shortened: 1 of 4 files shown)


src/Summary.cc
```diff
@@ -145,6 +145,22 @@ void Summary::readPool( const ResPool & pool )
     }
   }
 
+  if ( hasViewOption( PATCH_REBOOT_RULES ) ) {
+    // bsc#1183268: Patch reboot-needed flag overrules included packages.
+    // Work around the unfortunate case, where a maintenance patchs metadata
```
bzeller, line 150:
```suggestion
    // Work around the unfortunate case, where a maintenance patch's metadata
```

src/Summary.cc
```diff
@@ -145,6 +145,22 @@ void Summary::readPool( const ResPool & pool )
...
+    // do not request a reboot, while some included package does.
```
bzeller, line 151:
```suggestion
    // does not request a reboot, while some included package does.
```

src/Summary.cc
```diff
@@ -145,6 +145,22 @@ void Summary::readPool( const ResPool & pool )
...
+    // The 'patch --skip-interactive' command operates based on the maintenance
+    // metadata and skips patches that require a reboot. In this case we suppress
+    // the reboot needed hint of included packages in order not to confuse.
```
bzeller, line 154:
```suggestion
    // the reboot needed hint of included packages in order not to confuse the user.
```

bzeller (approved):
Ok but some suggestions for the big comment

Issue #388, opened 2021-05-24
[Feature request] Add comment to package lock


cbcoutinho:
Hi,

I have disabled a third-party package as it unnecessarily conflicts with an openSUSE package by using `zypper addlock`. To make this easier to understand for others on the system, I would like to add a comment to the lock, similar to how repositories can have aliases.

I was hoping for something like this:

```shell
$ sudo zypper addlock -r bad-repo --comment "Conflicts with packageA in good-repo" packageB
```

mlandres:
A good idea.

mlandres:
zypper 1.14.48 will support addlock *-m, --comment* to add a comment when specifying a lock.
Thanks to @coolgw who implemented it.

PR #393, opened 2021-06-12
Add comment option for lock command

coolgw:
fix https://github.com/openSUSE/zypper/issues/388


mlandres (changes requested):
Looks good so far. The new option should also be described in the man page (doc/zypper.8.txt).
(make builds the manpage out of the asciidoc file. `man -l FILE` can be used to check the page formatting)

coolgw:
> Looks good so far. The new option should also be described in the man page (doc/zypper.8.txt).
> (make builds the manpage out of the asciidoc file. `man -l FILE` can be used to check the page formatting)

Update doc/zypper.8.txt

mlandres (approved):
Thanks.

Issue #391, opened 2021-06-09
"..search --sort-by-name.." is case-insensitive


MrM40:
zypper version 1.14.45 (current 15.3 stable)
I would expect the following command would output a sorted package-list but that's not the case:
`zypper search --sort-by-name wind` 

(I am aware --sort-by-name is default)

Result:
![packages](https://user-images.githubusercontent.com/16258063/121428162-c8abf000-c975-11eb-985f-ed279d68a65d.jpg)

mlandres:
You mean, it should be sorted case-insensitive, right?

MrM40:
Yea, I just noticed that it myself! 
Then I guess it's less important / low priority.

PR #396, opened 2021-06-27
Case-insensitive sorting of search results (issue #391)

f-matt:
Case-insensitive sorting should be used for other columns as well? If so, converting column values to the same case, before performing the comparison, should be enough?

### DIFF
--- a/src/Table.h
+++ b/src/Table.h
@@ -258,7 +258,9 @@ public:
         } else
           return ( noL && ! noR ? -1 : ! noL && noR ?  1 : 0);
       }
-      return ( a_r._columns[curr_column_r] < b_r._columns[curr_column_r] ? -1 : a_r._columns[curr_column_r] > b_r._columns[curr_column_r] ?  1 : 0 );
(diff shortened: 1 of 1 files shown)


src/Table.h
```diff
@@ -258,7 +258,9 @@ class TableRow
         } else
           return ( noL && ! noR ? -1 : ! noL && noR ?  1 : 0);
       }
-      return ( a_r._columns[curr_column_r] < b_r._columns[curr_column_r] ? -1 : a_r._columns[curr_column_r] > b_r._columns[curr_column_r] ?  1 : 0 );
+      std::string lower_a_r = str::toLower(a_r._columns[curr_column_r]);
+      std::string lower_b_r = str::toLower(b_r._columns[curr_column_r]);
```
mlandres, line 262: Using `str::compareCI` (strcasecmp) would be the preferred way in such a  case because it avoids creating 2 string copies. But I'll close the PR because it will be solved by using new compare function, that will also ignore embedded ANSI SGR chars.

PR #404, opened 2021-07-21
Miscfixes

mlandres:


bzeller (approved):
Ok

src/Summary.cc
```diff
@@ -1667,15 +1667,22 @@ void Summary::writeXmlResolvableList( std::ostream & out, const KindToResPairSet
...
 {
+  Zypper & zypper( Zypper::instance() );
+
   unsigned pkgchanged = _inst_pkg_total;
   const auto & iter = _toremove.find( ResKind::package );
   if ( iter != _toremove.end() )
     pkgchanged += iter->second.size();
 
+  bool need_restart = ( _need_restart && zypper.runtimeData().plain_patch_command && !(_viewop & UPDATESTACK_ONLY) );
+  bool need_reboot = ( _need_reboot_patch || _need_reboot_nonpatch );
```
bzeller, line 1678: Small improvement could be to move those calculations into functions so that they are not duplicated

mlandres: done

bzeller (approved):

PR #411, opened 2021-08-31
Avoid calling 'su' to detect a too restrictive sudo user umask

mlandres:
[bsc#1186602](https://bugzilla.suse.com/show_bug.cgi?id=1186602).


bzeller (approved):
ok

Issue #401, opened 2021-07-16
allow to specify key id like --gpg-auto-import-key-id 1234asdf-1234asdf


JanZerebecki:
Allow to specify the key id like `--gpg-auto-import-key-id 1234asdf-1234asdf` to only trust that key and cryptographically verify it matches the given id/fingerprint. This would be much easier than:
```
wget ....key
gpg --list-packets ...key|head -n6|tail -n1|egrep '^\W+keyid: 1234asdf1234asdf'
gpg --import ....key
gpg --export 1234asdf1234asdf |rpmkeys --import -
```

dwymark:
I found that I needed to make some small adjustments to the example sketched here in order to import a single key manually into the rpm keyring.

The below is not working on my machine with RPM 4.17.0:
```
gpg --export 1234asdf1234asdf | rpmkeys --import -
```

I found that (1) rpmkeys doesn't seem to interpret `-` as stdout as expected, instead it throws a read error; and that (2) rpmkeys wants the key to import to be in ascii armor format.

```
gpg --armor --export 1234asdf1234asdf > a.pub
rpmkeys --import a.pub
```

The above works and allowed me to manually import a single key from my keyring into the rpm keyring.

dwymark:
I gave this a shot, since I became familiar with related parts of the code base while working on [issue 348](https://github.com/openSUSE/zypper/issues/348). My first cut is in the work-in-progress PR linked to above.

I have a couple questions before I finalize the PR.

1. Where are the log messages sent to `MIL` logged to? I would like to do a quick sanity check to make sure my log statements are being written as expected.
2. In my current implementation, I do not allow the "--gpg-auto-import-key-id" argument to be repeated. Should we allow users to specify multiple keys that they would accept by repeating this parameter?
3. I couldn't find any tests for `--gpg-auto-import-keys`. Are the behavior of these parameters included in any test suites? I would be happy to write tests for my contribution if desired.

mlandres:
>     1. Where are the log messages sent to `MIL` logged to? I would like to do a quick sanity check to make sure my log statements are being written as expected.

Executed as root: `/var/log/zypper.log`. Or redirected to stderr: `ZYPP_LOGFILE=- zypper ....` or a filename instead of the `-`.
(From within zypper usually as component `[zypper]` and loglevel `<1>`=MIL)

[...]

PR #421, opened 2021-11-13
Allow auto import only for specific key

dwymark:
Implemented a new command option "--gpg-auto-import-key-id" which allows users to only auto-import a GPG key with the provided key ID. This was done in response to a feature request in [issue 401](https://github.com/openSUSE/zypper/issues/401).

### DIFF
--- a/src/Config.cc
+++ b/src/Config.cc
@@ -297,6 +297,7 @@ Config::Config()
   , reboot_req_non_interactive( false )
   , no_gpg_checks( false )
   , gpg_auto_import_keys( false )
+  , gpg_auto_import_key_id( "" )
   , machine_readable( false )
   , no_refresh( false )
   , no_cd( false )
@@ -569,6 +570,18 @@ std::vector<ZyppFlags::CommandGroup> Config::cliOptions()
               })),
               // translators: --gpg-auto-import-keys
               _("Automatically trust and import new repository signing keys.")
+        },
+        { "gpg-auto-import-key-id", 0, ZyppFlags::RequiredArgument, std::move( ZyppFlags::StringType( &gpg_auto_import_key_id, "", "KEY_ID").
+            after( [](const ZyppFlags::CommandOption &, const boost::optional<std::string> & key_id) {
+              std::string warn = str::form(
+                  _("Turning on '%s' for key id '%s'. The specified signing key will be automatically imported if present!"),
+                  "--gpg-auto-import-key-id",
+                  key_id->c_str());
+              Zypper::instance().out().warning( warn, Out::HIGH );
+              MIL << "gpg-auto-import-key-id is on for " << *key_id << endl;
+            })),
+            // translators: --gpg-auto-import-key-id
+            _("Automatically trust and import the specified repository signing key, if it is present.")
         },
         { "plus-repo", 'p', ZyppFlags::Repeatable | ZyppFlags::RequiredArgument, ZyppFlags::GenericContainerType( plusRepoFromCLI, ARG_URI ),
               // translators: --plus-repo, -p <URI>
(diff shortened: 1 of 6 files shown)


src/Config.cc
```diff
@@ -570,6 +571,18 @@ std::vector<ZyppFlags::CommandGroup> Config::cliOptions()
               // translators: --gpg-auto-import-keys
               _("Automatically trust and import new repository signing keys.")
         },
+        { "gpg-auto-import-key-id", 0, ZyppFlags::RequiredArgument, std::move( ZyppFlags::StringType( &gpg_auto_import_key_id, "", "KEY_ID").
```
mlandres, line 574: It's IMO debatable whether a new option is introduced or `gpg-auto-import-keys` is amended to take an optional argument. In any case a key or `,`-separated list of keys should be accepted.

src/callbacks/keyring.h
```diff
@@ -273,8 +273,9 @@ namespace zypp
         // gpg key info
         dumpKeyInfo( s << std::endl, key_r, context_r )  << std::endl;
 
-        // if --gpg-auto-import-keys or --no-gpg-checks print info and don't ask
-        if (_gopts.gpg_auto_import_keys)
+        // if --gpg-auto-import-keys, --no-gpg-check, or --gpg-auto-import-key-id matches key, print info and don't ask.
+        if (_gopts.gpg_auto_import_keys
+            || (!_gopts.gpg_auto_import_key_id.empty() && _gopts.gpg_auto_import_key_id == key_r.id()))
```
mlandres, line 278: `_gopts.gpg_auto_import_key_id == key_r.id()` IMO is to strict.

The argument should be ID or FPR (recommended 16byte/40byte). As convenience a short-ID (8byte insecure) could be accepted. This would be `key_r.providesKey( _gopts.gpg_auto_import_key_id )`.

At least the manpage should state that the short-ID is considered to be insecure. Optionally the argparser could print a note/warning if a `! PublicKeyData::isSafeKeyId( _gopts.gpg_auto_import_key_id )` is passed.

src/Config.cc
dwymark: Great, I will make sure to accept comma separated keys as well.

I also like your suggestion of amending `gpg-auto-import-keys` rather than adding a new parameter. Do you want to open up that conversation to other stakeholders, or can we move forward with the single parameter implementation?

mlandres: We can move on. Staying with `gpg-auto-import-key` and amending it to 

1. take an optional argument (comma separated list of IDs)
2. allow multiple occurrences of  `gpg-auto-import-key`

should be feasible for the argparser .

This enables the user to use `--gpg-auto-import-key=A,B` or `--gpg-auto-import-key=A --gpg-auto-import-key=B` as they like it.

It also allows to solve the issue of `--gpg-auto-import-key  ... --gpg-auto-import-key=A` in a natural way. No errors about conflicting options are needed. The options build up a set of key IDs. An empty set accepts all keys as it currently does. Otherwise only the keys in the set are auto-accepted. Keys not in the set are treated as usual.

(@bzeller JFYI)

dwymark: In order to make `gpg-auto-import-keys` optional, while still accepting comma separated lists of IDs as you suggest in your comment here, I had to make a small adjustment to the GenericContainerType definition in the argparser. If there was a better way to go about achieving this, let me know and I would be happy to make adjustments if necessary.

The change to GenericContainerType is in [its own commit](https://github.com/openSUSE/zypper/pull/421/commits/e21b6e9ac136b734a78d16413cc9a31ed204fc98).

src/callbacks/keyring.h
dwymark: Change made in commit 180266c

PR #423, opened 2021-12-20
Translation Typo

Ryuno-Ki:
Remove errorneous t in German translation.

Yesterday, I switched to GeckoLinux ROLLING Mate and configured it for German locale.
Today, I noticed a small error in the help output of zypper and looked into submitting a patch.

I took a look at https://en.opensuse.org/openSUSE:Zypper_development but couldn't find a `trunk` branch.
According to https://github.com/openSUSE/zypper/pulls PRs seem to be opened against `master`, so I hope this is correct.
Looking at https://github.com/openSUSE/zypper/commits/master I can see something about Weblate used for translations, but I wasn't able to discover an URL, so please point me to it in case that's used instead.

Tagging @skriesch as she is the last one listed in `de.po` as localisation team. (as of https://github.com/openSUSE/zypper/pull/103 ?)

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -164,7 +164,7 @@ msgstr ""
 #: src/Config.cc:377
 msgid "User defined transaction id used in history and plugins."
 msgstr ""
-"tBenutzerdefinierte Transaktions-ID, die in der Historie und Plugins benutzt "
+"Benutzerdefinierte Transaktions-ID, die in der Historie und Plugins benutzt "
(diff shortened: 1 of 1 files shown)


mlandres:
@Ryuno-Ki  Danke.
JFYI: Translations are maintained at https://l10n.opensuse.org/projects/zypper/master/, but a patch for the .po file is fine as well.

PR #440, opened 2022-06-09
versioncmp: if verbose, also print the edition 'parts' whci are compared

mlandres:
```
> zypper -v vcmp 0.9.5+git20190908+3abfab2 0.9.5~git20210406.e554475

0.9.5+git20190908+3abfab2 is newer than 0.9.5~git20210406.e554475
LHSversion: : 0 : 9 : 5 : git : 20190908 :        3 : abfab :      2 : 
RHSversion: : 0 : 9 : 5 :   ~ :      git : 20210406 :     e : 554475 : 
```


src/commands/utils/versioncmp.cc
```diff
@@ -78,5 +78,68 @@ int VersionCompareCmd::execute( Zypper &zypper, const std::vector<std::string> &
...
+    };
+    std::vector<std::string_view> ls;
+    std::vector<std::string_view> rs;
+    std::string l { lhs.version() };
+    std::string r { rhs.version() };
+    auto str { zypper.out().info() };
+
+    split( l, ls, "LHSversion:" );
+    split( r, rs, "RHSversion:" );
+    print( str << "", ls, rs );
```
bzeller, line 133: Why is the code passing a `str << ""` instead of simply str?

src/commands/utils/versioncmp.cc
```diff
@@ -78,5 +78,68 @@ int VersionCompareCmd::execute( Zypper &zypper, const std::vector<std::string> &
...
+
+    l = lhs.release();
+    r = rhs.release();
+    if ( l.size() || r.size() ) {
+      split( l, ls, "LHSrelease:" );
+      split( r, rs, "RHSrelease:" );
+      print( str << "", ls, rs );
```
bzeller, line 140: Same as above, why passing a empty string into the stream

mlschroe:
You also need to support the handling of `^`.

mlandres:
> You also need to support the handling of `^`.

Right. I'll display them as well.

src/commands/utils/versioncmp.cc
```diff
@@ -78,5 +78,70 @@ int VersionCompareCmd::execute( Zypper &zypper, const std::vector<std::string> &
...
+      vs.clear();
+      vs.push_back( first );
+      for ( const char * s = v.c_str(); *s; ) {
+        unsigned cs = chclass( *s );
+        if ( cs == 0 ) {
+          ++s;
+          continue;
+        }
+        const char * e = s+1;
+        while ( chclass( *e ) == cs )
```
mlschroe, line 103: If think you change to add a `*e &&` to the while condition, otherwise you'll overrun the buffer if the version ends with a character of class 0.

mlschroe: Oh wait, ignore me ;-)

PR #441, opened 2022-06-14
Moved logrotate files from user specif directory /etc/logrotate.d to …

schubi2:
…vendor specif directory /usr/etc/logrotate.d.

So, users changes can still be done in /etc and will not be overwritten by an update.

Logrotate supports /usr/etc now. /etc should only be used for individual changes from the admin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,10 @@ INCLUDE(ZyppCommon)
 OPTION (ENABLE_BUILD_TRANS "Build translation files by default?" ON)
 OPTION (ENABLE_BUILD_TESTS "Build and run test suite by default?" OFF)
 
+#Vendor specific directory
+SET( DISTCONFDIR "/usr/etc")
+ADD_DEFINITIONS( -DDISTCONFDIR="${DISTCONFDIR}" )
+
 # VERSION
 INCLUDE( ${ZYPPER_SOURCE_DIR}/VERSION.cmake )
 SET( VERSION "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}" )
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ INCLUDE(ZyppCommon)
 OPTION (ENABLE_BUILD_TRANS "Build translation files by default?" ON)
 OPTION (ENABLE_BUILD_TESTS "Build and run test suite by default?" OFF)
 
+#Vendor specific directory
+SET( DISTCONFDIR "/usr/etc")
+ADD_DEFINITIONS( -DDISTCONFDIR="${DISTCONFDIR}" )
+
```
mlandres, line 19:
```suggestion
IF (DISTCONFDIR)
    MESSAGE(STATUS "DISTCONFDIR is set to ${DISTCONFDIR}")
ELSE (DISTCONFDIR)
    SET( DISTCONFDIR ${SYSCONFDIR} )
ENDIF (DISTCONFDIR)

```

zypper.spec.cmake
```diff
@@ -175,8 +176,9 @@ rm -rf "$RPM_BUILD_ROOT"
 %license COPYING
 %endif
 %config(noreplace) %{_sysconfdir}/zypp/zypper.conf
-%config(noreplace) %{_sysconfdir}/logrotate.d/zypper.lr
-%config(noreplace) %{_sysconfdir}/logrotate.d/zypp-refresh.lr
+%dir %{_distconfdir}/logrotate.d
+%{_distconfdir}/logrotate.d/zypper.lr
+%{_distconfdir}/logrotate.d/zypp-refresh.lr
```
mlandres, line 181: If `%{_distconfdir}` is unset or `== %{_sysconfdir}`, the files must go `%config(noreplace) %{_sysconfdir}/...` as before.

mlandres (changes requested):

PR #442, opened 2022-06-14
Add --download-max-retries and --download-retry-wait-time switches (bsc#1200370)

bzeller:


src/Zypper.cc
```diff
@@ -24,8 +24,8 @@
 
 #include <zypp/base/LogTools.h>
 #include <zypp/base/Algorithm.h>
-#include <zypp/base/UserRequestException.h>
-#include <zypp/base/DtorReset.h>
+#include <zypp-core/base/UserRequestException>
+#include <zypp-core/base/DtorReset>
```
mlandres, line 28: AFAIKS the new header file locations are in the buildrequired libzypp version. But if the new ZConfig variable is also used, the buildrequires must be adapted. Preferably in a single committ like 2eb7e57fd8057a42c3113afd05dc359ea7e22e12 and _before_ the new featues are used in the code.

mlandres (changes requested):

PR #462, opened 2022-10-20
Update cs.po with the mice instead of hedgehog

jankohoutek:
Apodemus uralensis cimrmani  is one and only endemic taxon vertebrate animal in the Czech republic, so it should be in the Czech translation instead of the hedgehog

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -2892,9 +2892,9 @@ msgid ""
 "  \\\\\\\\\\\\\\__o\n"
 "__\\\\\\\\\\\\\\'/_"
 msgstr ""
-"   \\\\\\\\\\\n"
-"  \\\\\\\\\\\\\\__o\n"
-"__\\\\\\\\\\\\\\'/_"
+"  (o)(o)--."
+"   \oo/ (  )"
+"   m\/m--m'`--------."
 
 #. translators: command synopsis; do not translate lowercase words
 #: src/commands/nullcommands.cc:37


mlandres:
```
# zypper moo
   \\\\\
  \\\\\\\__o
__\\\\\\\'/_
```
That's hard to judge for me. The `moo` command showing the hedgehog was implemented by @mvidner,  member of the Prague team, in 2006. AFAIK it is a kind of mascot. I can't change it without the teams approval.

mvidner (approved):
The team evolves, so the mascot can evolve too :) Go ahead

po/cs.po
```diff
@@ -2892,9 +2892,9 @@ msgid ""
 "  \\\\\\\\\\\\\\__o\n"
 "__\\\\\\\\\\\\\\'/_"
 msgstr ""
-"   \\\\\\\\\\\n"
-"  \\\\\\\\\\\\\\__o\n"
-"__\\\\\\\\\\\\\\'/_"
+"  (o)(o)--."
```
mlandres, line 2895:
```suggestion
"  (o)(o)--.\n"
```
1st line needs a NewLine char `\n` appended.

po/cs.po
```diff
@@ -2892,9 +2892,9 @@ msgid ""
...
+"   \oo/ (  )"
```
mlandres, line 2896:
```suggestion
"   \\oo/ (  )\n"
```
Inside a Cstring a literal BackSlash (\\) needs to be escaped `\\`.
2nd line needs a NewLine char `\n` appended.

po/cs.po
```diff
@@ -2892,9 +2892,9 @@ msgid ""
...
+"   m\/m--m'`--------."
```
mlandres, line 2897:
```suggestion
"   m\\/m--m'`--------."
```
Inside a Cstring a literal BackSlash (\\) needs to be escaped `\\`.
(No NewLine char on the last line.)"

mlandres (changes requested):
NewLines and literal BackSlashes need to be fixed.

PR #465, opened 2022-11-07
Zypper serviceadd fails on idempotency check

mlandres:
[bsc#1203715](https://bugzilla.opensuse.org/show_bug.cgi?id=1203715)
(Plus the man page update for `.no_auto_prune`)


bzeller (approved):
OK

PR #484, opened 2023-04-18
Correct typo.

pfee:


mlandres:
Thanks. (It's a translated string, so I included the .po files rather than just merging your PR)

PR #497, opened 2023-08-04
Changed location of bash-complication (bsc#1213854).

arozyev:
This PR changes the location of zypper.sh bash completion script, from /etc/bash_completion.d/ to /usr/share/bash-completion/completions/


src/CMakeLists.txt
```diff
@@ -259,7 +259,7 @@ INSTALL(
 # bash completion script
 INSTALL(
   FILES bash-completion.sh
-  DESTINATION ${SYSCONFDIR}/bash_completion.d
+  DESTINATION ${DATADIR}/bash-completion/completions
```
mlandres, line 262: After reading the [CMAKE doc](https://cmake.org/cmake/help/latest/command/install.html#installing-files) I'd suggest to simply use the builtin `${CMAKE_INSTALL_DATADIR}` and not passing the `DATADIR` down from the spec file.
The benefit is, that the install will work (with respect to the prefix set) even if no `-DDATADIR` is passed to cmake. I see, we pass -DMANDIR too, but it may very well be, that this is some legacy, and the CMake files could use ${CMAKE_INSTALL_MANDIR} as well.

arozyev: ok, thanks for the hint, done.

Issue #52, opened 2014-09-30
Search for ZyppCommon.cmake in the cmake config path of library "Zypp".


sunweaver:
Hi,

if the cmake config of libzypp (aka "Zypp.cmake") is installed at /usr/<lib>/cmake/Zypp/, we have to make sure that not only Zypp.cmake gets found by zypper's cmake configuration, but also the ZyppCommon.cmake file needs to be properly located.

This patch handles this:
http://anonscm.debian.org/cgit/collab-maint/zypper.git/tree/debian/patches/1001_make-ZyppCommon-cmake-module-includable.patch

PR #500, opened 2023-08-27
Find ZyppCommon on multi-arch systems without having to set ZYPP_PREFIX

bluca:
On multi-arch distributions such as Debian or Ubuntu we need to check a different path in case ZYPP_PREFIX is not set.

Fixes https://github.com/openSUSE/zypper/issues/52


bluca:
Gentle ping @mlandres - would be nice to fix this long-standing issue, it's the last patch we carry downstream in Debian.

bzeller (approved):
Looks good to me, all testcases still build fine so we can use that.

PR #501, opened 2023-08-27
Fix some typos and spelling errors found by Lintian

bluca:


mlandres:
Lintian should not change `msgid` lines in `.po` and `.pot` files. They are automatically extracted from the source files and usually not mergable.

Issue #504, opened 2023-09-24
Zypper info exit code


aeiplatform:
Hello,
Why is exit code for `zypper info non-existing-package` 0?

Every other package manager is not (apt, dnf, pacman, ...).
Is this intentional?

mlandres:
No, this is not intentional. It should behave like search and return 104.

PR #506, opened 2023-10-06
Fix issue #504

Chlorophytus:
Hi, I'm doing some code stuff for Hacktoberfest and I figured I'd start off with this.

I am not sure how to go about this, but I kind of know C++.

I think this is one approach of getting Zypper to return exit code 104 if something isn't found in the `zypper info` command.

Let me know what else I need to do.

### DIFF
--- a/src/commands/query/info.cc
+++ b/src/commands/query/info.cc
@@ -127,7 +127,7 @@ int InfoCmd::execute( Zypper &zypper, const std::vector<std::string> &positional
   }
 
   printInfo( zypper, positionalArgs_r, _options );
-  return ZYPPER_EXIT_OK;
+  return zypper.exitInfoCode();
 }
 
 std::string InfoCmd::summary() const
--- a/src/commands/query/info.cc
+++ b/src/commands/query/info.cc
@@ -127,7 +127,7 @@ int InfoCmd::execute( Zypper &zypper, const std::vector<std::string> &positional
   }
 
   printInfo( zypper, positionalArgs_r, _options );
-  return zypper.exitInfoCode();
+  return ZYPPER_EXIT_OK;
 }
 
 std::string InfoCmd::summary() const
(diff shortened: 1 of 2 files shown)


src/info.cc
```diff
@@ -240,6 +240,7 @@ void printInfo(Zypper & zypper, const std::vector<std::string> &names_r, const P
 
     if ( q.empty() )
     {
+      zypper.setExitInfoCode( ZYPPER_EXIT_INF_CAP_NOT_FOUND );
```
mlandres, line 243: I'd set the ExitInfo a bit later, if the fallback query is empty or no fallback is wanted (the two `continue` cases).

src/commands/query/info.cc
```diff
@@ -127,7 +127,7 @@ int InfoCmd::execute( Zypper &zypper, const std::vector<std::string> &positional
   }
 
   printInfo( zypper, positionalArgs_r, _options );
-  return ZYPPER_EXIT_OK;
+  return zypper.exitInfoCode();
```
mlandres, line 130: Actually this should not be needed. [`main()`](https://github.com/openSUSE/zypper/blob/master/src/main.cc#L162) checks whether zypper would return EXIT_OK, and if it would it returns any pending ExitInfo code instead. Internally  ExitInfo should not be treated as an error nor should it disguise any _real_ errors. That's why they are handles at the end.

mlandres (changes requested):

src/info.cc
Chlorophytus: Good thinking, I might've tried that but I'll try moving the `setExitInfoCode` call again.

src/commands/query/info.cc
Chlorophytus: Fixed.

mlandres (approved):
Thanks.

PR #518, opened 2023-11-20
Moving /etc/zypp/zypper.conf to /usr/etc/zypp/zypper.conf

schubi2:
/etc/zypp/zypper.conf will be used for admin changed values only.


schubi2:
Michl will implement his own solution.

PR #521, opened 2023-12-07
patch: Make sure reboot-needed is remembered until next boot

mlandres:
[bsc#1217873](https://bugzilla.suse.com/show_bug.cgi?id=1217873)

### DIFF
--- a/src/solve-commit.cc
+++ b/src/solve-commit.cc
@@ -990,6 +990,11 @@ void solve_and_commit ( Zypper &zypper, SolveAndCommitPolicy policy )
             zypper.setExitCode( ZYPPER_EXIT_INF_REBOOT_NEEDED );
             zypper.out().warning(_("One of the installed patches requires a reboot of"
             " your machine. Reboot as soon as possible."), Out::QUIET );
+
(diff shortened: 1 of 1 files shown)


bzeller (commented):
Why are we doing this in zypper and not in libzypp? Shouldn't Yast also benefit from the fix?

mlandres:
At the end of the day because of maintenance requested to let their patches overrule the package settings in  `zypper patch`. This is solely a hack for the `patch` command. The fact that a patch as a whole is flagged as `needs-reboot` has it's downsides because the patch describes not just the fixed package but also dependent packages.  And even more because the patch status is a computed state.

- Installing just a dependent package which happens to be also mentioned in a patch may turn the patch to `applied ` but does IMO not justify to set `needs-reboot` if the package did not carry it. In the summary it would be hard to explain why installing or updating just a doc package requires a reboot.

- Pure package operations may change the computed state of a bunch of patches although the patches have not been part of the transaction. Installing a new package may turn all patches ever released and mentioning it into `not-needed'. If this package version dos not claim to need a reboot, why should libzypp flag the system, just because some old patch had this flag. This is most relevant after packages have been split or joined because the packages current content determines the need to reboot, not it's ancient one.

PR #527, opened 2024-01-24
make_obs_url fix platform for newer Leap releases

lkocman:
We no longer use openSUSE_Leap_15.6 for Leap repositories. Nowadays it's just 15.6 (in fact since 15.3 I believe).

```
lkocman@localhost:~> sudo zypper ar obs://devel:tools:ide:vscode devel_tools_ide_vscode
[sudo] password for root: 
Guessed: platform = openSUSE_Leap_$releasever
Adding repository 'devel_tools_ide_vscode' ................................................................................................................................................................[done]
Repository 'devel_tools_ide_vscode' successfully added

URI         : https://download.opensuse.org/repositories/devel:/tools:/ide:/vscode/openSUSE_Leap_15.6
Enabled     : Yes
GPG Check   : Yes
Autorefresh : No
Priority    : 99 (default priority)

Repository priorities are without effect. All enabled repositories share the same priority.

```

The url should be https://download.opensuse.org/repositories/devel:/tools:/ide:/vscode/15.6

### DIFF
--- a/src/utils/misc.cc
+++ b/src/utils/misc.cc
@@ -569,8 +569,8 @@ Url make_obs_url( const std::string & obsuri )
         }
         else if ( platform == "Leap" && pdata.vendor() == "openSUSE" )
         {
-          // bsc#1187425 Hotfix
-          platform = "openSUSE_Leap_$releasever";
+          // openSUSE Leap 15.3+ uses simply "15.X" as a reference to the codestream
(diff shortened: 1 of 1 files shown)


lkocman:
I suppose we do not care about older releases which are EOL nowadays. 15.4 was marked as EOL on Dec 31 2023.

PR #529, opened 2024-01-30
Don't print reboot required message when only downloading

jubalh:
Since we are only downloading and not installing there is no reboot required.

### DIFF
--- a/src/Summary.cc
+++ b/src/Summary.cc
@@ -1635,7 +1635,7 @@ void Summary::dumpTo( std::ostream & out )
     // patch command (auto)restricted to update stack patches
     Zypper::instance().out().notePar( 4, _("Package manager restart required. (Run this command once again after the update stack got updated)") );
   }
-  if ( showNeedRebootHInt() )
+  if ( showNeedRebootHInt() && !_download_only)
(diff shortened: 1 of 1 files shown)


jubalh:
@mlandres  ping

mlandres:
Sorry, I forgot to add a comment. 

I see the point, but for other messages - like `package manager restart needed`  - the same could be true. But more important that many people use a combination of `--dry-run --downlod-only` because this also performs the fileconflict check after downloading. And with dry-run you may want to see what would happen if... even if it's not performed now. 

That's why I added a line before the continue prompt which tells whether some of the _fake_ options are in use:
```
Note: System reboot required.

Backend:  classic_rpmtrans --download-only
Continue? [y/n/v/...? shows all options] (y): 
```
We could perhaps also refine the note rather than suppressing it.

jubalh:
So should I not just check whether `--dry-run` is used?

mlandres:
No. I'll change the notification texts to reflect that this is what would be needed if download-only/dry-run were not active.

This way one can see what would be needed if the transaction would be applied and it should be clear that it does not apply to the current command. I think that's better than hiding the message.

jubalh:
Alright!

PR #538, opened 2024-03-25
New option --remove-orphaned to remove all orphaned packages in dup.

mlandres:
[bsc#1221525](https://bugzilla.suse.com/show_bug.cgi?id=1221525)

### DIFF
--- a/doc/zypper.8.txt
+++ b/doc/zypper.8.txt
@@ -736,6 +736,9 @@ It might be less tedious to persistently set *$releasever* to the target distrib
 	*--from* _alias_|_name_|_#_|_URI_::
 		The option can be used multiple times and restricts the upgrade to the specified repositories only. Nevertheless all enabled repositories are visible to the resolver and will be considered to satisfy dependency problems.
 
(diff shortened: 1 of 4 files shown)


bzeller (approved):

karlmistelberger:
Nice option. Many users ask for `--remove-unneeded`: https://forums.opensuse.org/search?q=zypper%20packages%20--unneeded%20order%3Alatest They currently use something like:
```
zypper packages --unneeded | grep "^i" | cut -d"|" -f3 | xargs zypper remove --clean-deps
```

mlandres:
There will appear a `zypper remove --remove-unneeded` in #540

ivan-mingolov-blue-technologies:
> There will appear a `zypper remove --remove-unneeded` in #540

Is that still the case, because I can't find such a change in #540 or you're planning to expose such an option in future PR?

mlandres:
@ivan-mingolov-blue-technologies It will appear in a future PR. We got interrupted and releases what we dis so far.

PR #540, opened 2024-03-27
Misc command fixes

mlandres:


mlandres:
@bzeller a couple of independent commits.

Important is the last one fixing [bsc#1222086](https://bugzilla.suse.com/show_bug.cgi?id=1222086). It's basically done. Just formating the _Time since last refresh_ needs to be done (ATM just _NUM s_)

src/repos.cc
```diff
@@ -712,9 +717,22 @@ int repo_specs_to_aliases( Zypper & zypper, const std::vector<std::string> & rsp
...
+  std::optional<PropertyTable> mdstats;
+  if ( geteuid() != 0 ) {
+    MIL << "turn off refresh for not-root user" << endl;
+    zypper.configNoConst().no_refresh = true;
+    now = Date::now();
+    mdstats = PropertyTable();
+    mdstats->add( _("Repository"),
+                  // translator: used as a table header
+                  _("Time since last refresh") );
+  }
```
bzeller, line 734: Shouldn't we also check if a chroot is used instead of / ? A normal user might be able to refresh it

bzeller:
```Repository                                                  : Time since last refresh
SUSE Internal CA Certificate (openSUSE_Tumbleweed_20240409) : 3107504 s
code                                                        : 59839 s
Main Repository (NON-OSS)                                   : 59838 s
Main Repository (OSS)                                       : 59838 s
Main Update Repository                                      : 59837 s
google-chrome                                               : 59837 s
multimedia                                                  : 59837 s
packman                                                     : 59837 s
openSUSE-Tumbleweed-Debug                                   : 59836 s
Open H.264 Codec (openSUSE Tumbleweed)                      : 59836 s
openSUSE-Tumbleweed-Source                                  : 59835 s
```

Can we show those in minutes/hours depending on how long the last refresh was ago? Seconds is growing a bit big

bzeller:
Also should the `--userinstalled` and `--autoinstalled` switches be in `search` also? Might be a non existing usecase but i could see all `lib*` packages that are `autoinstalled`

src/repos.cc
mlandres: Good point.

mlandres:
> Can we show those in minutes/hours depending on how long the last refresh was ago? Seconds is growing a bit big

That's "formating the Time since last refresh needs to be done"...

mlandres:
> Also should the `--userinstalled` and `--autoinstalled` switches be in `search` also? 

We may do this as well. The lists in `packages` were just low hanging.

bzeller (approved):

PR #544, opened 2024-04-25
Fix const reference typo

dirkmueller:


mlandres:
@dirkmueller Wrong branch? src/Table.h no longer has code.

mlandres:
And libzypp(zypp-tui/Table.h) has ` for ( const SortParam &sortParam : _by_columns ) {`

dirkmueller:
I was apparently on a older code level of that branch.

PR #547, opened 2024-04-30
openSUSE Weblate merge conflict resolved

MeggyCal:
The username-password auth was disabled, so I was unable to find a way to resolve it as a `opensuse-i18n` user. Hope it looks good.


mlandres:
Thanks @MeggyCal

PR #555, opened 2024-06-24
Fixing minor grammar issue in summary output

rdoxenham:
In the zypper summary output when a package (or packages) are being added, it summarises, but does so with a minor grammatical error, where the word "an" is missing before "additional". Incredibly minor, but it bugs me :)


rdoxenham:
Based on the recent commits, this is no longer required.

PR #556, opened 2024-07-01
Fix readline setup to handle Ctrl-C and Ctrl-D corrrectly (bsc#1227205)

mlandres:


src/main.cc
```diff
@@ -34,6 +34,35 @@ namespace env
...
+  fd_set rfds;
+  struct timeval tv;
+  Zypper & zypper( Zypper::instance( true ) );
+  while ( true ) {
+    FD_ZERO( &rfds );
+    FD_SET( fileno(f), &rfds );
+    tv.tv_sec  = 0;
+    tv.tv_usec = 500000;
+
+    int retval = select( 1, &rfds, NULL, NULL, &tv );
```
bzeller, line 48: replace with g_poll or poll , select can have problems with fd's > 1024

bzeller (changes requested):

PR #571, opened 2024-10-12
Update Zypper.cc

smith518:
Fix: Prevent multiple simultaneous downloads for the same file after metalink redirect.


mlandres:
Thanks, but - apart from the fact that zypper does not do any downloads at all - you want to add 2 functions which never get called and not even compile?

PR #578, opened 2024-11-19
RCF: Feat: add autoremove

Lunarequest:
This is a request for comment on a zypper autoremove command. As a system in updated during its life cycle packages can get left behind. While zypper has `rm --cleap-deps` and `dup --remove-orphaned` both have short comings. rm only handles the package passed to it and removing orphans can remove packages a user explicitly built and installed themselves.

To bring zypper inline with the wider linux package manager system(dnf, apt/apt-get) I added a zypper autoremove command. This is code does require reworking of some part but as it stands this will clean up packages unneeded and orphaned packages based on flags passed.

Thanks @janvhs who helped debug some issues with the code


afaerber:
It looks like RCF in the title meant RFC? If you don't want it to be merged yet, you should mark it as draft, which prevents that. Otherwise just remove RFC from title and description, in case maintainers don't require changes and would like to merge it.

Lunarequest:
> It looks like RCF in the title meant RFC? If you don't want it to be merged yet, you should mark it as draft, which prevents that. Otherwise just remove RFC from title and description, in case maintainers don't require changes and would like to merge it.

ah thanks I didn't realize that I typoed RFC, I'll move it to a draft for now, I believe it should be fine to just merge it, in the current state.

src/Command.cc
```diff
@@ -24,6 +24,7 @@
 #include "commands/query.h"
 #include "commands/installremove.h"
 #include "commands/sourceinstall.h"
+#include  "commands/autoremove.h"
```
afaerber, line 27: Extra space? or tab?

src/commands/autoremove.h
```diff
@@ -0,0 +1,37 @@
...
+#include "commands/solveroptionset.h"
+
+class AutoRemoveCmd : public ZypperBaseCommand
+{
+public:
+    AutoRemoveCmd(std::vector<std::string> &&commandAliases_r);
+
+private:
+    bool _details = false;
+    bool _orpahned = false;
```
afaerber, line 21: Typo: _orphaned

src/global-settings.h
```diff
@@ -92,6 +92,12 @@ struct DupSettingsData
 };
 using DupSettings = GlobalSettingSingleton<DupSettingsData>;
 
+struct ArmSettingsData
```
afaerber, line 95: I'm biased here, but would encourage a less confusing naming than "Arm" if feasible.

src/commands/autoremove.cc
```diff
@@ -0,0 +1,121 @@
...
+        _("You are run autoremove on the system. This is experimental and subject to change"
+          " This may remove unintended packages please confirm the packages that will be removed before you"
+          " continue. See '%s' for more information about this command."),
+        "man zypper"));
+
+    std::vector<std::string> pkgs;
+
+    auto checkStatus = [=](const PoolItem &pi_r) -> bool
+    {
+        // isipi : prevent returning true if identical installed items are tested.
```
afaerber, line 77: "isipi"?

afaerber (changes requested):
Appreciate this Hackweek project!
Some minor nits pointed out inline.

src/commands/autoremove.cc
```diff
@@ -0,0 +1,121 @@
...
+#include "src/SolverRequester.h"
+#include "commonflags.h"
+#include "Zypper.h"
+
+extern ZYpp::Ptr God;
+
+AutoRemoveCmd::AutoRemoveCmd(std::vector<std::string> &&commandAliases_r) : ZypperBaseCommand(std::move(commandAliases_r),
+                                                                                              // translators: command synopsis; do not translate lowercase words
+                                                                                              _("autoremove (arm) [OPTIONS]"),
+                                                                                              // translators: command summary: dist-upgrade, dup
```
afaerber, line 22: Copy&paste? "command summary: autoremove, arm"?

Lunarequest: ah copy and past my bad

mlandres:
In fact a `rm --cleap-deps` without arguments is WIP together with commands to allow manipulating the auto-installed status of packages. The later is needed to be able to tune the `clean-deps` result because no user-requested packages should be removed here. Also `--cleap-deps` in install/update commands to clean in the dependency tree of update packages. A dedicated autoremove command should not be needed or become just an alias for `rm --cleap-deps`.

The kind of downside of your approach is that it is zypper only and that it's not under direct control of the resolver. I.e. you pass explicit remove requests for packages based on a previous computation rather than asking the resolver itself to compute the job. Implemented in libzypp, the job will become available to YAST and PK as well.

PR #595, opened 2025-02-11
Let zypper dup fail in case of (temporarily) unaccessible repos

mlandres:
(bsc#1228434, bsc#1236939, fixes#446)

### DIFF
--- a/src/commands/basecommand.cc
+++ b/src/commands/basecommand.cc
@@ -175,6 +175,15 @@ int ZypperBaseCommand::defaultSystemSetup( Zypper &zypper, SetupSystemFlags flag
     init_repos( zypper );
     if ( zypper.exitCode() != ZYPPER_EXIT_OK )
       return zypper.exitCode();
+    else if ( flags_r.testFlag( FailIfReposFail ) && zypper.exitInfoCode() != ZYPPER_EXIT_OK ) {
+      // Fail as demanded by FailIfReposFail. Commands may tune or explain
+      // this by overriding systemSetup (e.g. distupgrade).
+      // NOTE: Since several exitInfoCodes are documented to be returned as a kind
+      // of annotation, if the command would otherwise have returned 0, we indicate
+      // a real error here. Just in case the command does not amend it later.
+      zypper.setExitCode( ZYPPER_EXIT_ERR_ZYPP ); // intentionally an error and NOT the exitInfoCode
+      return zypper.exitInfoCode();               // but return the exitInfoCode
+    }
   }
 
   DtorReset _tmp( zypper.configNoConst().disable_system_resolvables );
--- a/src/commands/distupgrade.cc
+++ b/src/commands/distupgrade.cc
@@ -44,6 +44,22 @@ zypp::ZyppFlags::CommandGroup DistUpgradeCmd::cmdOptions() const
   });
 }
 
+int DistUpgradeCmd::systemSetup( Zypper &zypper )
+{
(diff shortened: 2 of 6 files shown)


src/commands/basecommand.cc
```diff
@@ -175,6 +175,13 @@ int ZypperBaseCommand::defaultSystemSetup( Zypper &zypper, SetupSystemFlags flag
     init_repos( zypper );
     if ( zypper.exitCode() != ZYPPER_EXIT_OK )
       return zypper.exitCode();
+    else if ( flags_r.testFlag( FailIfReposFail ) && zypper.exitInfoCode() != ZYPPER_EXIT_OK ) {
+      zypper.out().taggedPar( 4, MSG_ERRORString( command()[0]+":" ), _("Due to the treatment of orphaned packages dist-upgrade depends on a proper repository setup more than any other command. It must not continue if enabled repositories fail to refresh. This may severely damage the system. If a failing repository is actually not needed, it must be disabled.")
+      +std::string(" ")
+      +_("See 'man zypper' for more information about this command.") );
```
bzeller, line 181: Should this be really inside basecommand? The wording of the error message suggests that is a error which can happen only in dup.  This could be completely handled in DistUpgradeCmd::systemSetup. Since its overloaded already a simple if() checking the return code of defaultSystemSetup should be enough?

bzeller: Also shouldn't this check for ZYPPER_EXIT_INF_REPOS_SKIPPED rather then  != ZYPPER_EXIT_OK ?

src/commands/distupgrade.cc
```diff
@@ -44,6 +44,14 @@ zypp::ZyppFlags::CommandGroup DistUpgradeCmd::cmdOptions() const
   });
 }
 
+int DistUpgradeCmd::systemSetup( Zypper &zypper )
+{
+  // bsc#1228434, bsc#1236939, bsc#961719, bsc#961724, et.al.
+  // dup must not continue if repositories failed to refresh.
+  SetupSystemFlags flags = DefaultSetup | FailIfReposFail;
+  return defaultSystemSetup( zypper, flags );
```
bzeller, line 52: Error message should be here since its a dup only error

src/repos.cc
```diff
@@ -934,10 +934,12 @@ void do_init_repos( Zypper & zypper )
...
     zypper.out().error(_("Some of the repositories have not been refreshed because of an error.") );
-    // TODO: A user abort during repo refresh as well as unavailable metadata
-    // should probably lead to ZYPPER_EXIT_ERR_ZYPP right here. Ignored refresh
-    // errors may continue. For now at least remember the refresh error to prevent
-    // a 0 exit code after the action completed. (bsc#961719, bsc#961724, et.al.)
+    // A user abort during repo refresh as well as unavailable metadata
+    // could lead to ZYPPER_EXIT_ERR_ZYPP right here. Ignored refresh
+    // errors may continue. For now at least remember the refresh error
+    // to prevent a 0 exit code after the action completed.
+    // In ZypperBaseCommand::defaultSystemSetup commands may decide whether
```
bzeller, line 941:
```suggestion
    // Commands should override ZypperBaseCommand::systemSetup to decide whether
```

bzeller (changes requested):

mlandres:
Move the specific code into distupgrade. Fixed the comments as suggested.

bzeller (approved):

PR #609, opened 2025-04-03
Use libzypp improvements for preload and mirror handling

bzeller:


mlandres:
I'd remove `Use RepoInfo::effectiveBaseUrls() when needed ` . Commands like `zypper lr` should stay with showing the baseurl and not one of the mirrorurls. (For this RepoInfo::url() in openSUSE/libzypp#628 needs to be reverted as well).

PR #622, opened 2025-08-13
Fixed `bash-completion`: `zypper refresh` now ignores repository priorities

bberberov:
In the completion for `zypper refresh`, the `sed` filter was not excluding the "Repository priorities in effect:" from `zypper -q lr`.


src/bash-completion.sh
```diff
@@ -140,7 +140,7 @@ _zypper() {
 			addservice | as | modifyservice | ms | removeservice | rs)
 				opts=(${opts[@]}$(echo; LC_ALL=POSIX $ZYPPER -q ls | \
 					sed -rn '/^[[:blank:]]*[0-9]/{
-						s/^[[:blank:]]*[0-9]+[[:blank:]]*\|[[:blank:]]*([^|]+).*/\1/
+						s/^[[:blank:]]*[0-9]+[[:blank:]]+\|[[:blank:]]+([^|]+).*/\1/
```
mlandres, line 143: Please leave them as `[[:blank:]]*` please, here and in the other hunks. It should not rely on a pretty printed output.

mlandres:
Thank you.

PR #643, opened 2026-04-14
Add --filter-version-change to zypper lu

l-mb:
Adds filtering by version change significance to reduce noise in update listings. Supports levels: rebuild (hides rebuild-only changes) and package (hides all release-only changes).

*Rationale*: I've often wanted to see just changes that are more likely to change actual code, not "mere" rebuilds triggered by dependencies. I've previously filtered this via a script from zypper lu output, but I think this is cleaner.

Local testing suggests this is working fine, and the code looks clean enough. I made a guess at an acceptable name for the option.

### DIFF
--- a/doc/zypper.8.txt
+++ b/doc/zypper.8.txt
@@ -548,6 +548,15 @@ If *patch* is specified, zypper acts as if the *list-patches* command was execut
 	*--best-effort*::
 		See the *update* command for description.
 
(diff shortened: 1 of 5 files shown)


mlandres (approved):
Danke schön!

mlandres:
zypper-1.14.97

